PR #4, opened 2015-10-27
commands for adding and removing modules

simkam:


Ladicek:
Looks good on the first sight. I'll have a look at it next week again and merge, I don't anticipate any big changes. In the meantime, could you please get rid of the `@author` javadoc tags? You can extract this info from Git already :-)

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/modules/AddRemoveModuleTest.java
```diff
@@ -0,0 +1,120 @@
...
+                .property("foo", "bar")
+                .property("john", "doe")
+                .build();
+        client.apply(addModule);
+
+        // verify that module was added
+        File asRoot = new File(getPathToAs());
+
+        File module = new File(asRoot, "modules" + File.separator + TEST_MODULE_NAME.replaceAll("\\.", File.separator));
+        assertTrue("Module " + module.getAbsolutePath() + " doesn't exist on path", module.exists());
```
Ladicek: I always found this style of assertion failure messages confusing. I'd prefer the "should" style; in this case, `Module ... should exist on path`.

commands/src/main/java/org/wildfly/extras/creaper/commands/modules/AddModule.java
```diff
@@ -0,0 +1,170 @@
...
+
+        if (moduleXml != null)
+            cmd.append(" --module-xml=" + moduleXml);
+
+        if (mainClass != null)
+            cmd.append(" --main-class=" + mainClass);
+
+        Joiner joiner = Joiner.on(',');
+        if (!resources.isEmpty()) {
+            cmd.append(" --resources=" + joiner.join(resources));
```
Ladicek: What happens if the path to the resource contains spaces (and/or quotes)? See also the `PatchingConversions` class. There should probably be some kind of a builder for local CLI operations that would encapsulate that, but for the time being, you have to solve that manually.

Ladicek:
BTW, please also rebase on master.

simkam:
* fixed resource-delimiter
* assertion failure messages changed to 'should' style
* spaces in resources paths are now escaped 
  * I think that quotes in path are so rare that is no need to handle them specially, also `Builder` accepts resource as `File` only, so I don't expect that anybody mistakenly adds path encapsulated in quotes 
* added path with spaces to test

commands/src/main/java/org/wildfly/extras/creaper/commands/modules/AddModule.java
```diff
@@ -0,0 +1,171 @@
...
+
+        if (moduleXml != null)
+            cmd.append(" --module-xml=" + moduleXml);
+
+        if (mainClass != null)
+            cmd.append(" --main-class=" + mainClass);
+
+        Joiner resourcesJoiner = Joiner.on(resourceDelimiter);
+        if (!resources.isEmpty()) {
+            cmd.append(" --resources=" + resourcesJoiner.join(resources).replaceAll(" ", "\\\\ "));
```
Ladicek: What if the resource delimiter is a space here? It probably can't be, but that's not documented anywhere. I mean -- I'm fine with throwing an exception if an assumption is violated, but if there are such assumptions, they should be checked (and documented).

commands/src/main/java/org/wildfly/extras/creaper/commands/modules/AddModule.java
```diff
@@ -0,0 +1,171 @@
...
+        ctx.client.executeCli(cmd.toString());
+    }
+
+    @Override
+    public String toString() {
+        return "AddModule " + moduleName;
+    }
+
+    public static final class Builder {
+        public static final String DEFAULT_SLOT = "main";
```
Ladicek: I'd probably move this to the top-level `AddModule` class, because I think that a builder shouldn't hold constants related to the class it builds, but I'll leave that up to you.

I was thinking about making it `private`/`protected`/default, but I agree that it makes sense as `public`.

Ladicek:
Let's fix that string escaping thing, otherwise LGTM.

simkam:
* added validation of values in Builder
* changed `resourceDelimiter` type to `char`
* default slot constant moved `AddModule` class

commands/src/main/java/org/wildfly/extras/creaper/commands/modules/AddModule.java
```diff
@@ -0,0 +1,197 @@
...
+         * @param moduleName the name of the module to be added
+         */
+        public Builder(String moduleName) {
+            this(moduleName, DEFAULT_SLOT);
+        }
+
+        /**
+         * @param moduleName the name of the module to be added
+         * @param slot       specifies a slot which should be created. If this argument is not specified, "main" slot
+         *                   is assumed.slot of the new module
```
Ladicek: `is assumed.slot of the new module`? :-)

commands/src/main/java/org/wildfly/extras/creaper/commands/modules/RemoveModule.java
```diff
@@ -0,0 +1,50 @@
...
+ * as well as its directory structure up to the point where other modules met.</p>
+ */
+public final class RemoveModule implements OnlineCommand {
+    private final String moduleName;
+    private final String slot;
+
+    /**
+     * @param moduleName the name of the module to be removed
+     * @param slot       specifies a slot which should be removed. If this argument is not specified, "main" slot
+     *                   is assumed.slot of the new module
```
Ladicek: `is assumed.slot of the new module`? :-)

Ladicek:
Besides the two tiny issues, the test fails with EAP 6 with this exception: `org.jboss.as.cli.CommandFormatException: Unrecognized arguments: [--resource-delimiter]`. Not really sure about this... do we even need that option?

simkam:
updated
* fixed javadoc
* --resource-delimiter is ignored if `serverVersion` < 2.0.0  (WF8)

test still fail with EAP6, it doesn't accept paths with spaces. What to do? 

all fails with EAP6
module add --name=test --resources=/tmp/path with spaces/jboss-negotiation-extras.jar
module add --name=test --resources=/tmp/path\ with\ spaces/jboss-negotiation-extras.jar
module add --name=test --resources="/tmp/path\ with\ spaces/jboss-negotiation-extras.jar"
module add --name=test --resources="/tmp/path with spaces/jboss-negotiation-extras.jar"

commands/src/main/java/org/wildfly/extras/creaper/commands/modules/AddModule.java
```diff
@@ -0,0 +1,204 @@
...
+
+        if (moduleXml != null)
+            cmd.append(" --module-xml=" + moduleXml);
+
+        if (mainClass != null)
+            cmd.append(" --main-class=" + mainClass);
+
+        Joiner resourcesJoiner = Joiner.on(File.pathSeparatorChar);
+        // resource-delimiter was added in WF 8, WFLY-1871
+        if (ctx.serverVersion.greaterThan(ManagementVersion.VERSION_2_0_0)) {
```
Ladicek: Should this be `greaterThanOrEqualTo`?

simkam: fixed

simkam:
* category `WildFlyTests` added to test
* version check fixed

PR #24, opened 2015-11-27
Added socket binding command

honza-kasik:
A command for creating socket binding.

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBinding.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBinding.java
@@ -1,0 +1,176 @@
+package org.wildfly.extras.creaper.commands.socketbindings;
+
+import org.wildfly.extras.creaper.core.CommandFailedException;
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * <p>A command for creating socket binding under default socket binding group if not specified else.</p>
+ * <p>This command <b>does not check</b> whether the port or multicast port are available!</p>
+ */
+public final class AddSocketBinding implements OnlineCommand {
+
+    private final String socketBindingName;
+    private final String socketBindingGroup;
+
+    private final List<String> clientMappings;
+    private final Boolean fixedPort;
+    private final String interfaceName;
+    private final String multicastAddress;
+    private final Integer multicastPort;
+    private final Integer port;
+    private final boolean replaceExisting;
+
+    private AddSocketBinding(Builder builder) {
+        this.socketBindingName = builder.socketBindingName;
+        this.socketBindingGroup = builder.socketBindingGroup;
+        this.clientMappings = builder.clientMappings;
+        this.fixedPort = builder.fixedPort;
+        this.interfaceName = builder.interfaceName;
+        this.multicastAddress = builder.multicastAddress;
+        this.multicastPort = builder.multicastPort;
+        this.port = builder.port;
+        this.replaceExisting = builder.replaceExisting;
+    }
+
+    @Override
+    public void apply(OnlineCommandContext ctx) throws CommandFailedException, IOException {
+        Operations ops = new Operations(ctx.client);
+
+        String socketBindingGroup = this.socketBindingGroup;
+        if (socketBindingGroup == null) {
+            socketBindingGroup = ctx.client.options().isDomain ? "full-sockets" : "standard-sockets";
+        }
+
+        Address socketBindingAddress = Address.of("socket-binding-group", socketBindingGroup)
+                .and("socket-binding", socketBindingName);
+
+        if (replaceExisting) {
+            try {
+                ops.removeIfExists(socketBindingAddress);
+            } catch (OperationException e) {
+                throw new CommandFailedException("Failed to remove existing socket binding " + socketBindingName, e);
+            }
+        }
+
+        ops.add(socketBindingAddress, Values.empty()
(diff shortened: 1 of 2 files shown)


commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBinding.java
```diff
@@ -0,0 +1,174 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * <p>A command for creating socket binding under default socket binding group if not specified else.</p>
+ * <p>This command <b>does not check</b> whether the port or multicast port are available!</p>
+ *
+ * @author Jan Kasik <jkasik@redhat.com>
```
Ladicek: Please remove the `@author` tag, this information is available from Git history.

commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBinding.java
```diff
@@ -0,0 +1,174 @@
...
+ */
+public class AddSocketBinding implements OnlineCommand {
```
Ladicek: Could this be `final`?

honza-kasik: I think it could and it is now.

commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBinding.java
```diff
@@ -0,0 +1,174 @@
...
+    public void apply(OnlineCommandContext ctx) throws CommandFailedException, IOException {
+        Operations ops = new Operations(ctx.client);
+        String contextSocketBindingGroup;
+
+        if (socketBindingGroup == null) {
+            contextSocketBindingGroup = ctx.client.options().isDomain ? "full-sockets" : "standard-sockets";
+        } else {
+            contextSocketBindingGroup = socketBindingGroup;
+        }
+
```
Ladicek: I think I would prefer something like this to inventing a new name `context...`:

```
String socketBindingGroup = this.socketBindingGroup;
if (socketBindingGroup == null) {
    ...
}
```

Also, your choice of default socket binding group needs a bit of explanation. It's trivial for standalone, but not so much for managed domain. I guess that `full-sockets` is chosen because it's the socket binding group used by all the servers that are started in the default domain configuration (i.e., `server-one` and `server-two` from `main-server-group`; there's `server-three` from `other-server-group`, but that's not started by default). It kinda makes sense, but should probably be mentioned in the class javadoc.

Ladicek:
Looks good. A command for removing a socket binding is probably not _that_ useful, but you surely could add a test ;-)

commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBinding.java
Ladicek: Or possibly not in the class javadoc, but the javadoc of the `Builder` constructor that doesn't take the socket binding group name.

honza-kasik: I added the javadoc and I took your advice regarding avoiding inventing new names :)

Ladicek:
Add a test, squash the commits and I'll merge.

Ladicek:
In addition, please also rebase on current master.

honza-kasik:
Added test, squashed and rebased.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBindingTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.admin.Administration;
+
+import java.io.IOException;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * @author Jan Kasik <jkasik@redhat.com>
+ *         Created on 1.12.15.
+ */
```
Ladicek: Please no.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBindingTest.java
```diff
@@ -0,0 +1,61 @@
...
+@RunWith(Arquillian.class)
+public class AddSocketBindingTest {
+
+    private String SOCKET_BINDING_NAME = "creaper-sb";
+    private String SOCKET_BINDING_GROUP = "standard-sockets";
+    private Address TEST_SOCKET_BINDING_ADDRESS = Address.of("socket-binding-group", SOCKET_BINDING_GROUP)
+            .and("socket-binding", SOCKET_BINDING_NAME);
```
Ladicek: Should these constants be `static final`?

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBindingTest.java
```diff
@@ -0,0 +1,61 @@
...
+            client.close();
+        }
+    }
+
+    @Test
+    public void addSocketBinding() throws CommandFailedException, IOException, OperationException {
+        AddSocketBinding addSocketBinding = new AddSocketBinding.Builder(SOCKET_BINDING_NAME)
+                .build();
+        client.apply(addSocketBinding);
+        Assert.assertTrue(ops.exists(TEST_SOCKET_BINDING_ADDRESS));
```
Ladicek: Hah, I was thinking that maybe we should test for a port number here, but that's error-prone. So OK, this will do.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBindingTest.java
honza-kasik: Made it static final, thanks

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/socketbindings/AddSocketBindingTest.java
honza-kasik: Sorry, removed.

PR #25, opened 2015-11-30
Support mcp attribute with data sources

jmartisk:


Ladicek:
Looks good. Is it easy to add a test (or rather modify an existing one), or would that be too scary?

ochaloup:
And maybe it would be have this for offline command as well.

Ladicek:
+1, I think that generally online and offline variants should be symmetric.

jmartisk:
Fixed.

Ladicek:
What's that commit message? :-) Otherwise, LGTM.

jmartisk:
Fixed the 'fixed' keyword :)

Ladicek:
Thanks!

PR #41, opened 2015-12-16
avoid exceptions when Operations.removeIfExists is called from command

### LINKED ISSUE #40: exception thrown when calling Operations.removeIfExists from command
When calling `Operations.removeIfExists` outside of an `OnlineCommand`, it returns `false` correctly when the desired management resource doesn't exist. However, when it's called from an `OnlineCommand`, it throws an exception. The reason is that inside a command, the management client is wrapped in a proxy that detects failed management operations and throws exceptions automatically, so that commands don't have to do that.

This is an important convenience that needs to be retained. However, sometimes it's simply expected that an operation will fail (not only for `Operations.removeIfExists` but also for `Operations.exists` and one could argue that this can happen for user-constructed operations as well). So there needs to be a way to communicate such intent from inside the command. If we were on Java 8, a lambda would be natural, but on Java 6, this is simply too unwieldy.

Thanks to @olukas for discovering the bug!

Ladicek:
When calling `Operations.removeIfExists` outside of an `OnlineCommand`, it returns `false` correctly when the desired management resource doesn't exist. However, when it's called from an `OnlineCommand`, it throws an exception. The reason is that inside a command, the management client is wrapped in a proxy that detects failed management operations and throws exceptions automatically, so that commands don't have to do that.

This is an important convenience that needs to be retained. However, sometimes it's simply expected that an operation will fail (`Operations.removeIfExists` is one case, `Operations.exists` is another and one could argue that this can happen for user-constructed operations as well). So there needs to be a way to communicate such intent from inside the command. If we were on Java 8, a lambda would be natural, but on Java 6, this is simply too unwieldy. Instead, a new method `allowFailures` is added to `OnlineManagementClient` that returns a `Closeable`. Calling that method and closing its result delimit the block of code in which failures shouldn't be converted to exceptions because they are expected. This is still somewhat wordy in Java 6, but in Java 7 and above, try-with-resources can be used to bring the syntactic overhead down.

Fixes #40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
   when `replaceExisting()` is used
 - the test suite now uses latest WildFly 10 pre-release (was blocked
   by wrong patching test)
+- fixed a failure when `Operations.batch` was passed an empty `Batch`
+- added `OnlineManagementClient.expectedFailures` to avoid exceptions
+  when operations are executed from commands and failures are expected
+  (e.g. `Operations.exists` or `Operations.removeIfExists`)
 
 ## 0.9.2
 
--- a/core/src/main/java/org/wildfly/extras/creaper/core/offline/OfflineCommand.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/offline/OfflineCommand.java
@@ -7,7 +7,7 @@ import org.wildfly.extras.creaper.core.Command;
  * A command is typically comprised of multiple management operations and is <b>not</b> transactional. It is expected
  * that there will be <b>no</b> other process manipulating the same configuration file.</p>
  *
- * <p>Error handling for commands is provided automatically by the {@code ManagementClient}. That is, if you are
+ * <p>Error handling for commands is provided automatically by the {@code ManagementClient}. That is, when you are
  * implementing the {@code apply} method, you <i>should</i> let all the exceptions bubble up. Only catch exceptions
  * if you can recover from them. This is signified by the {@code throws} clause of the {@code apply} method anyway.</p>
  */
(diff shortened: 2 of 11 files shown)


core/src/main/java/org/wildfly/extras/creaper/core/online/ExpectedFailuresBlock.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * <pre>
+ * try (ExpectedFailuresBlock expectedFailures = client.expectedFailures()) {
+ *     ModelNodeResult result = client.execute(...);
+ *     ...
+ * }
+ * </pre>
+ */
+public interface ExpectedFailuresBlock extends Closeable {
+}
```
rhatlapa: I would suggest different name as you don't expect failures, it just can occur and you don't want the command to fail because of it.

What about naming it rather something like `AllowFailuresBlock`? WDYT?

Ladicek: Yea, I'm not particularly fond of the name. It's really meant for a very specific situation. I chose `ExpectedFailures` because outside of such blocks, failures are _not expected_, but I agree that a there should be a better name.

rhatlapa:
I like the whole change, just I would prefer naming the method and class using **allow** rather than **expect** as in such blocks the failures are not expected, they can just occur and you just don't want for the commands to automatically fail because of it.

Ladicek:
I've been thinking for a while, but I didn't find anything better than "allow". So allow it is, thanks.

PR #53, opened 2016-01-22
Added commands for adding acceptors and connectors in messaging subsystem

honza-kasik:


Ladicek:
The `AddAcceptor` and `AddConnector` names are a bit too generic. There's a multitude of connectors in the applicatoin server. Can we figure out a better name?

Also, how often does one need to use these commands? They seem quite specific to me.

honza-kasik:
What about `AddMessagingAcceptor` and `AddMessagingConnector`?
I really have no idea how often these commands are needed. I just though that it would be nice to have them since I used in in several test cases...

Ladicek:
+1 to those names. They're unique and, at the same time, not that longer.

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
```diff
@@ -0,0 +1,230 @@
...
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * An abstract class which implements common properties of connector and acceptor in messaging connections.
```
Ladicek: You don't have to say that it's an abstract class, everyone can see that from `abstract class ...` below.

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
```diff
@@ -0,0 +1,230 @@
...
+    protected boolean replaceExisting;
+
+    //in-vm
+    protected Integer serverId;
+
+    //generic and remote
+    protected String socketBinding;
+
+    //remote
+    protected String factoryClass;
```
Ladicek: How many of the fields above can be made `final`? Ideally all of them, which would require slightly restructuring the construction logic, but some of them can be made `final` with no other changes.

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
```diff
@@ -0,0 +1,230 @@
...
+        this.name = builder.name;
+        this.serverName = builder.serverName;
+        this.params = builder.params;
+        this.replaceExisting = builder.replaceExisting;
+    }
+
+    /**
+     * Get type of transport configuration's instance
+     */
+    protected abstract TransportConfigItem getConfigType();
```
Ladicek: Instead of an `abstract` method, this should be a `final` field initialized in a constructor. Subclasses can pass the correct value to the superclass constructor just as easily as overriding a method.

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
```diff
@@ -0,0 +1,230 @@
...
+        protected InVmBuilder(String name, String serverName) {
+            super(name, serverName);
+        }
+
+        public THIS serverId(Integer serverId) {
+            this.serverId = serverId;
+            return (THIS) this;
+        }
+
+        public THIS validate() {
```
Ladicek: Should this really be `public`?

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AddAcceptor.java
```diff
@@ -0,0 +1,72 @@
...
+            super(name);
+        }
+
+        public RemoteBuilder(String name, String serverName) {
+            super(name, serverName);
+        }
+
+        @Override
+        public AddAcceptor build() {
+            return new AddAcceptor(this);
```
Ladicek: In `InVmBuilder`, you `return new AddAcceptor(validate())`, but here, you `return new AddAcceptor(this)`. How come? (My suggestion would be to make `validate` return `void` and call it in all `build` methods before the `return` statement.)

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/messaging/AddAcceptorOnlineTest.java
```diff
@@ -0,0 +1,94 @@
...
+                        .and("in-vm-acceptor", TEST_ACCEPTOR));
+        client.close();
+    }
+
+    @Test
+    public void addInVmAcceptor() throws CommandFailedException, IOException {
+        int serverId = 145872154;
+
+        AddAcceptor addAcceptor = new AddAcceptor
+                .InVmBuilder(TEST_ACCEPTOR)
```
Ladicek: I can sympathise with what you're trying to do, but this is definitely not idiomatic. Keep the entire FQN of the class on a single line: `new AddAcceptor.InVmBuilder(TEST_ACCEPTOR)`.

Ladicek:
Good work on test coverage, by the way!

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
honza-kasik: I tried to think about restructuring the construction logic to make fields `final`, but I cannot think out of the box anymore and I am stuck :). What do you suggest? Can you give me a hint, please?

Ladicek: Not critical, really. You could create a single `private` constructor that would take all the parameters and set the fields; all the other `protected` constructors would just delegate to this. But as I said, I don't insist on this; I just like the commands to be immutable.

honza-kasik:
Thank you! I added suggested improvements and fixed it.

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
```diff
@@ -0,0 +1,252 @@
...
+    protected final Integer serverId;
+
+    //generic and remote
+    protected final String socketBinding;
+
+    //remote
+    protected final String factoryClass;
+
+
+    public AbstractTransportConfigAddCommand(TransportConfigType configType, TransportConfigItem configItem,
```
Ladicek: `private`

Ladicek:
I think this looks good now, with that single comment about privacy of a constructor.

Ladicek:
Oh, and one more thing: the `validate` method in the builders should have a return type of `void`. The return value is never used anyway.

Ladicek:
OK, please squash and I'll merge.

honza-kasik:
Commits were squashed

Ladicek:
Why two commits? Are the tests in the 2nd commit somehow different from the tests in the 1st commit? Also, why the note "Fixed bugs and added suggested improvements" in the first commit?

honza-kasik:
Sorry, fixed.

honza-kasik:
Sorry, fixed

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
```diff
@@ -0,0 +1,249 @@
...
+        this.name = name;
+        this.serverName = serverName;
+        this.params = params;
+        this.replaceExisting = replaceExisting;
+        this.serverId = serverId;
+        this.socketBinding = socketBinding;
+        this.factoryClass = factoryClass;
+    }
+
+    protected AbstractTransportConfigAddCommand(InVmBuilder builder, TransportConfigItem configItem) {
```
zebrik: Can you parametrize reference to generic type with wildcard (i.e. `InVmBuilder<?> builder`) to prevent in this case IMO uneccessary compiler complaints?

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
```diff
@@ -0,0 +1,249 @@
...
+                builder.name,
+                builder.serverName,
+                builder.params,
+                builder.replaceExisting,
+                builder.serverId,
+                null,
+                null);
+    }
+
+    protected AbstractTransportConfigAddCommand(GenericBuilder builder, TransportConfigItem configItem) {
```
zebrik: `GenericBuilder<?> builder`

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
```diff
@@ -0,0 +1,249 @@
...
+                builder.name,
+                builder.serverName,
+                builder.params,
+                builder.replaceExisting,
+                null,
+                builder.socketBinding,
+                builder.factoryClass);
+    }
+
+    protected AbstractTransportConfigAddCommand(RemoteBuilder builder, TransportConfigItem configItem) {
```
zebrik: RemoteBuilder<?> builder

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
honza-kasik: Sure I can, I just did not understand why to generify this before. :+1:

honza-kasik:
Parametrized contructor arguments with wildcard.

commands/src/main/java/org/wildfly/extras/creaper/commands/messaging/AbstractTransportConfigAddCommand.java
Ladicek: We already have quite a bunch of unchecked conversions in `commands`. They are practically unavoidable when using generics to simulate self types. I don't mind much, really. @honza-kasik it's your choice; in fact, I was just getting to merge this when @zebrik added these comments :-)

Ladicek: Hah, took me too long to comment :-) Nevermind.

Ladicek:
I still don't understand why we have `AddAcceptorOnlineTest`, `AddConnectorOnlineTest`, `AddMessagingAcceptorOnlineTest` and `AddMessagingConnectorOnlineTest`.

honza-kasik:
I am really sorry, as I was squashing commits I totally overlooked this...

PR #55, opened 2016-02-01
Added commands for security subsystem

olukas:


Ladicek:
Not assigning a milestone just yet, this might take long and I don't want to block any particular release with this.

Ladicek:
On the first sight, there's quite a lot of standalone builder classes. Typically, builders are embedded in the class of the command they are building. Any reason why not to do that?

Also, see the `Logging` class, which is a bit more friendly entrypoint for the commands in the `logging` subsystem -- could be useful here as well.

olukas:
These standalone builders only extends common builder for adding login module. They serve as facade for AddLoginModule.Builder. Only reason for them is simplifying usage of AddLoginModule command for some commonly used login modules. They can be wrapped to own command classes, but since they serve as facade then these commands will not added any new functionality or something. They will become just an empty wrappers. I am not sure if it will be better to wrap them to that command classes. WDYT?

I have also taken a look on Logging class. It is nice entry point to subsystem. I can add similar entry point to Security subsystem.

Ladicek:
If the standalone builders just simplify usage of the "generic" builder and don't add any functionality, maybe these builders could just be folded to the `Security` entrypoint class?

olukas:
They can be folded to `Security` entrypoint class. However some inconveniences occur:

1) `Security` class will include a lot of lines which are not directly related to "orientation" in security subsystem (which is IMHO one of main reason for that class) - but OK, it is probably only aesthetic detail.
2) Adding a new builder will result to modifying `Security` class instead of simply creating of the new builder class.

If you think these inconveniences are insignificant I will move those builders to `Security` class.

commands/src/main/java/org/wildfly/extras/creaper/commands/security/AddAuthorizationModule.java
```diff
@@ -0,0 +1,168 @@
...
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+/**
+ * Add a new authorization classic policy module to given security domain.
+ */
+public class AddAuthorizationModule implements OnlineCommand, OfflineCommand {
```
Ladicek: `final`?

commands/src/main/java/org/wildfly/extras/creaper/commands/security/AddAuthorizationModule.java
```diff
@@ -0,0 +1,168 @@
...
+         * In case when you use this constructor then authorization module name is the same as its code.
+         */
+        public Builder(String code) {
+            this(code, code);
+        }
+
+        public Builder(String code, String name) {
+            if (code == null) {
+                throw new IllegalArgumentException("Code of the authorization module must be specified as non null "
+                        + "value");
```
Ladicek: No need to do this, long strings are fairly common and the checkstyle configuration accounts for them, so they shouldn't trigger a warning.

commands/src/main/java/org/wildfly/extras/creaper/commands/security/AddLoginModule.java
```diff
@@ -0,0 +1,167 @@
...
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+/**
+ * Add a new authentication classic login module to given security domain.
+ */
+public class AddLoginModule implements OnlineCommand, OfflineCommand {
```
Ladicek: `final`?

commands/src/main/java/org/wildfly/extras/creaper/commands/security/AddLoginModule.java
```diff
@@ -0,0 +1,167 @@
...
+        ops.add(loginModuleAddress, Values.empty()
+                .andOptional("code", code)
+                .andOptional("flag", flag)
+                .andOptional("module", module)
+                .andObjectOptional("module-options", Values.fromMap(moduleOptions))
+        );
+    }
+
+    @Override
+    public final void apply(OfflineCommandContext occ) throws CommandFailedException, IOException {
```
Ladicek: I wonder why do you call it `occ` :-) Not that it's bad, but I was hoping to establish a convention of `ctx`.

commands/src/main/java/org/wildfly/extras/creaper/commands/security/AddMappingModule.java
```diff
@@ -0,0 +1,166 @@
...
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+/**
+ * Add a new mapping classic mapping module to given security domain.
+ */
+public class AddMappingModule implements OnlineCommand, OfflineCommand {
```
Ladicek: `final`?

commands/src/main/java/org/wildfly/extras/creaper/commands/security/AddSecurityDomain.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+            this.securityDomainName = securityDomainName;
+        }
+
+        public Builder cacheType(String cacheType) {
+            this.cacheType = cacheType;
+            return this;
+        }
+
+        public Builder replaceExisting() {
```
Ladicek: Must document that it can cause server reload. See commands for XA datasources, that's a similar case.

commands/src/main/java/org/wildfly/extras/creaper/commands/security/RemoveMappingModule.java
```diff
@@ -0,0 +1,56 @@
...
+        ops.remove(Address.subsystem("security")
+                .and("security-domain", securityDomainName)
+                .and("mapping", "classic")
+                .and("mapping-module", mappingModuleName));
+    }
+
+    @Override
+    public void apply(OfflineCommandContext occ) throws Exception {
+        occ.client.apply(GroovyXmlTransform
+                .of(RemoveMappingModule.class)
```
Ladicek: Please don't split `GroovyXmlTransform.of`.

Ladicek:
Comments are minor, this is a really good work. Does it work with EAP 6 as well?

Ladicek:
OK, I really like this in the current form. I found two small issues:

The class `AddLoginModule` should mention the builders in the class javadoc.

The `Remove*OnlineTest` tests have ugly copy&paste errors. For example, from `RemoveLoginModuleOnlineTest`:

```java
    @Test
    public void removeOneOfMoreLoginModules() throws Exception {
        client.apply(ADD_LOGIN_MODULE_1);
        assertTrue("The login module should be created", ops.exists(TEST_LOGIN_MODULE1_ADDRESS));
        client.apply(ADD_LOGIN_MODULE_2);
        assertTrue("The login module should be created", ops.exists(TEST_LOGIN_MODULE2_ADDRESS));

        client.apply(new RemoveLoginModule(TEST_SECURITY_DOMAIN_NAME, LOGIN_MODULE1_CODE));
        assertFalse("The security domain should be created", ops.exists(TEST_LOGIN_MODULE1_ADDRESS));
    }
```

The assert message on the last line should reflect that this is about _login modules_, not security domains, and about _removing_, not creating. So: `"The login module should be removed"`.

And things like this are all over the `Remove*OnlineTest` tests.

But these are pretty small issues and I believe we don't have to have an extra round of review on that. So when you fix it, please also rebase and squash. I'll happily merge.

Ladicek:
One other thing: the `AddLoginModule` builders use generics to emulate self types. With the datasources builders, I added a comment `The {@code THIS} type parameter is only meant to be used by subclasses. If you're not inheriting from this class, don't use it.` -- could you add it to the `AddLoginModule.Builder` javadoc too?

PR #56, opened 2016-02-01
add ManagementVersion 4.1.0

Ladicek:


Ladicek:
CC @mchoma

Ladicek:
Added a couple of small commits that are not worth their own PR.

Issue #42, opened 2015-12-17
Create online commands for deploying/undeploying


rhatlapa:
There should be created command, which will allow to easily create deployment operation, which can be used for both: server in standalone and in domain mode.

rhatlapa:
I would like to discuss how to create undeploy command for domain (its behaviour)

There is mainly one concern and that is deployment in repository vs being deployed to some specific server groups.

The deployment can't be removed from deployment repository till it is used by some server group.

Here is the problematic part, which is to decide whether it should be only undeployed/removed from specified server group or also from the deployment repository.

rhatlapa:
The easiest way seems to me making builder over CLI operation `undeploy` in this case. Where it would allow to specify whether it should be undeployed from all relevant server groups or just from specific ones.

Ladicek:
The undeploy command can have options specific to managed domain that would specify the intended behavior.

The key here is the default behavior where no such options are provided -- should it undeploy from all server groups and remove from content repository? Seems quite brutal, but probably closest to the standalone case. Or any other options?

Ladicek:
Fixed by #57.

Issue #10, opened 2015-11-05
Builders in child AddDatasources commands should return the child builder not the one from parent


ochaloup:
Currently when I create a `AddDatasource` command for a database - let's say `AddOracleDatasource` and I decide to use some of builder methods from parent (e.g. `addConnectionProperty`) then such method returns by fluent api the class of `AddDatasource.Builder` and not class of the child builder (e.g. `AddOracleDatasource.Builder`).

We need to apply generics magic (I'm sorry not the magic but F-bounded type polymorphism science) to get the builder from the child in our flow api.

We will need two generic parameters - one of type of the Builder and the second one of type the result command.

Ladicek:
Tentatively assigning to myself, but anyone should feel free to take this :-)

Ladicek:
I realized yesterday that we actually only need one type parameter -- the type of the builder itself. The `build` methods already return the correct type.

PR #57, opened 2016-02-01
Deploy command created.

rhatlapa:
Created deploy command, which handles deploy part of #42.

**EDIT**: updated link to point to correct issue.

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
@@ -1,0 +1,182 @@
+package org.wildfly.extras.creaper.commands.deployments;
+
+import org.jboss.as.controller.client.OperationBuilder;
+import org.jboss.dmr.ModelNode;
+import org.wildfly.extras.creaper.core.online.Constants;
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOY;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ENABLED;
+
+/**
+ * Command which takes care about deploying specified deployment under specified deployment name.
+ * <p/>
+ * In case of domain it is deployed by default to all server groups
+ */
+public final class Deploy implements OnlineCommand {
+
+    private final String deploymentName;
+    private final InputStream deploymentInputStream;
+    private final boolean autoCloseInputStream;
+    private final List<String> serverGroups;
+
+    private Deploy(InputStream deploymentInputStream, Builder deployCmdOptions) {
+        this.deploymentInputStream = deploymentInputStream;
+        this.deploymentName = deployCmdOptions.deploymentName;
+        this.autoCloseInputStream = deployCmdOptions.autoCloseInputStream;
+        this.serverGroups = deployCmdOptions.serverGroups;
+    }
+
+    @Override
+    public void apply(OnlineCommandContext ctx) throws Exception {
+        Operations ops = new Operations(ctx.client);
+
+        // deploy and add to server groups
+        ModelNode content = new ModelNode();
+        content.get(Constants.INPUT_STREAM_INDEX).set(0);
+
+        ModelNode deployRequest;
+
+        if (ctx.options.isDomain) {
+            List<String> groups = serverGroups;
+            if (serverGroups == null)
+                groups = ops.readChildrenNames(Address.root(), Constants.SERVER_GROUP)
+                        .stringListValue(Collections.<String>emptyList());
+
+            List<ModelNode> groupDeploymentAddresses = new ArrayList<ModelNode>(groups.size());
+            for (String serverGroup : groups) {
+                ModelNode mn = new ModelNode();
+                mn.add(Constants.SERVER_GROUP, serverGroup);
+                mn.add(Constants.DEPLOYMENT, deploymentName);
+                groupDeploymentAddresses.add(mn);
+            }
+            deployRequest = createDeploymentOperation(content,
+                    groupDeploymentAddresses.toArray(new ModelNode[groupDeploymentAddresses.size()]));
+        } else {
+            deployRequest = createDeploymentOperation(content);
+            ModelNode deployStep = deployRequest.get(Constants.STEPS).add();
+            ModelNode deploymentAddress = new ModelNode();
+            deploymentAddress.add(Constants.DEPLOYMENT, deploymentName);
+            deployStep.set(getEmptyOperation(DEPLOY, deploymentAddress));
+        }
+
+        OperationBuilder builder = new OperationBuilder(deployRequest, autoCloseInputStream);
+        builder.addInputStream(deploymentInputStream);
+        ctx.client.execute(builder.build());
+    }
+
+    private ModelNode createDeploymentOperation(ModelNode content, ModelNode... serverGroupAddresses) {
+        ModelNode composite = getEmptyOperation(Constants.COMPOSITE, new ModelNode().setEmptyList());
+        ModelNode steps = composite.get(Constants.STEPS);
+        ModelNode step1 = steps.add();
+        step1.set(getEmptyOperation(Constants.ADD, new ModelNode().add(Constants.DEPLOYMENT, deploymentName)));
+        step1.get(Constants.CONTENT).add(content);
+        if (serverGroupAddresses != null) {
+            for (ModelNode serverGroup : serverGroupAddresses) {
+                ModelNode sg = steps.add();
+                sg.set(getEmptyOperation(Constants.ADD, serverGroup));
+                sg.get(ENABLED).set(true);
+            }
+        }
+        return composite;
+    }
+
+    private ModelNode getEmptyOperation(String operationName, ModelNode address) {
+        ModelNode op = new ModelNode();
+        op.get(Constants.OP).set(operationName);
+        if (address != null) {
+            op.get(Constants.OP_ADDR).set(address);
+        } else {
+            // Just establish the standard structure; caller can fill in address later
+            op.get(Constants.OP_ADDR);
+        }
+        return op;
+    }
+
+
+    /**
+     * Builder for creating command for deploying deployment to the server.
+     */
+    public static final class Builder {
+        private File deploymentFile;
+        private InputStream deploymentInputStream;
+        private final boolean autoCloseInputStream;
+        private final String deploymentName;
+        private List<String> serverGroups;
+
+        public Builder(File deploymentFile) {
+            if (!deploymentFile.exists()) {
+                throw new IllegalArgumentException(deploymentFile + " doesn't exist!");
+            }
+            this.deploymentFile = deploymentFile;
+            this.deploymentName = deploymentFile.getName();
+            this.autoCloseInputStream = true;
+        }
+
+        public Builder(InputStream deploymentInputStream, String deploymentName,
+                       boolean autoCloseInputStream) {
+            this.deploymentInputStream = deploymentInputStream;
+            this.deploymentName = deploymentName;
+            this.autoCloseInputStream = autoCloseInputStream;
+        }
+
+        public Deploy build() {
+            InputStream inputStream = deploymentInputStream;
+            if (deploymentInputStream == null && deploymentFile != null) {
+                try {
+                    inputStream = new FileInputStream(deploymentFile);
+                } catch (FileNotFoundException ex) {
+                    throw new IllegalArgumentException(ex);
+                }
+            }
+            return new Deploy(inputStream, this);
+        }
+
+
+        /**
+         * <p>Domain mode only</p>
+         * <p>
+         *     Adds list of server groups for which the deployment should be applied.
+         * </p>
+         * <p>
+         *     Note you should use either {@code toAllServerGroup} or {@code toServerGroups}, not both.
+         * </p>
+         */
+        public Builder toServerGroups(String... serverGroups) {
+            if (this.serverGroups == null && serverGroups != null) {
+                this.serverGroups = new ArrayList<String>();
+            }
+            if (serverGroups != null) {
+                this.serverGroups.addAll(Arrays.asList(serverGroups));
+            }
+            return this;
+        }
+
+        /**
+         * <p>Domain mode only</p>
+         * <p>
+         *     Specifies that the deployment should be deployed to all server groups
+         * </p>
+         * <p>
+         *     Note you should use either {@code toAllServerGroup} or {@code toServerGroups}, not both.
+         * </p>
+         */
+        public Builder toAllServerGroups() {
+            this.serverGroups = null;
+            return this;
+        }
+
+    }
+}
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Undeploy.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Undeploy.java
@@ -1,0 +1,56 @@
+package org.wildfly.extras.creaper.commands.deployments;
+
+
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+
+/**
+ * Command which takes care about undeploying specified application (the content is by default also removed)
+ * <p>
+ * In case of domain the deployment is by default undeployed from all relevant server groups
+ * (server groups having the deployment enabled)
+ * </p>
+ */
+public final class Undeploy implements OnlineCommand {
+    private final String deploymentName;
+    private final boolean keepContent;
+
+    private Undeploy(Builder builder) {
+        this.deploymentName = builder.deploymentName;
+        this.keepContent = builder.keepContent;
+    }
+
+    @Override
+    public void apply(OnlineCommandContext ctx) throws Exception {
+        StringBuilder cmd = new StringBuilder("undeploy ").append(deploymentName);
+        if (keepContent) {
+            cmd.append(" --keep-content");
+        }
+        if (ctx.options.isDomain) {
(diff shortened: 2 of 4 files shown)


Ladicek:
How is #10 related? Should be #42, no?

Ladicek:
An option -- CLI has `deploy` which could be used here. I'm of course by no means objecting to creating the mgmt operation manually, just suggesting a possibility.

rhatlapa:
Yes, you are right, I meant #42

rhatlapa:
I wanted to use management operation directly to prevent need of creating temporary files and to just leave it in memory.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployCmdTest.java
```diff
@@ -0,0 +1,88 @@
...
+import org.wildfly.extras.creaper.test.WildFlyTests;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This test uses the {@code ajp} socket binding for all listener types, because there's no listener configured
+ * by default that uses it (unlike {@code http}).
+ */
```
Ladicek: Does it? :-)

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
```diff
@@ -0,0 +1,206 @@
...
+ * In case of domain it is deployed by default to all server groups
+ */
+public final class Deploy implements OnlineCommand {
+
+    private final String deploymentName;
+    private final InputStream deploymentInputStream;
+    private final boolean autoCloseInputStream;
+    private final List<String> serverGroups;
+
+    private Deploy(InputStream deploymentInputStream, Data deployCmdOptions) {
```
Ladicek: This is slightly worrying, because `InputStream` is a one-off thing. If we were on Java 8, I'd prefer `Supplier<InputStream>`. Another option would be Guava's `ByteSource`, and even though we already depend on Guava, I'd like to avoid depending on it in public APIs. Another very similar thing is ShrinkWrap's `Asset`. Sadly, on Java 6, there's probably nothing we can do (short of providing our own interface, which I'd really like to avoid) :-(

rhatlapa: Agree with you, don't now about any better way with Java 6 than using the InputStream. If that would be big issue, I would redo this and allow only File. In that case only in memory handling would be lost :-(.

Ladicek: No need to do that, we can live with `InputStream` for now.

Ladicek:
What are your plans with this? Do you want this reviewed and merged, or do you plan to work on undeploy commands first and prefer to keep this PR open?

rhatlapa:
I plan to work on undeploy command, I believe I should have time to do it during this week. If no one is in hurry, I would wait for it :-).

rhatlapa:
I have added also the undeploy command, now it fixes #42 as whole :-).

Note I was thinking also about removing all options from undeploy command, that it would in case of domain always undeploy the deployment from all relevant server groups. As the way the undeploy command behaves, it requires using enabling option for keeping content if there wasn't all relevant server groups selected. WDYT?

rhatlapa:
I've decided to remove the option for specifying server groups as I believe no one really wants to use it this way. If it is found needed, I left it as builder => it should be easy to add those options without breaking API.

rhatlapa:
=> ready for review

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
```diff
@@ -0,0 +1,211 @@
...
+            // Just establish the standard structure; caller can fill in address later
+            op.get(Constants.OP_ADDR);
+        }
+        return op;
+    }
+
+    /**
+     * A single shared instance of this class is passed through the entire builder invocation chain to accumulate
+     * all the options.
+     */
```
Ladicek: This is clearly copied from `Online|OfflineOptions`, where I deliberately broke the builder to a bunch of classes to be able to provide compile-time checking of all mandatory options. It adds quite some maintenance burden, so I don't think it's suitable for commands. Did you have any special reason to use this pattern instead of plain old single-class builder?

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
```diff
@@ -0,0 +1,211 @@
...
+
+            List<ModelNode> groupDeploymentAddresses = new ArrayList<ModelNode>(groups.size());
+            for (String serverGroup : groups) {
+                ModelNode mn = new ModelNode();
+                mn.add(Constants.SERVER_GROUP, serverGroup);
+                mn.add(Constants.DEPLOYMENT, deploymentName);
+                groupDeploymentAddresses.add(mn);
+            }
+            deployRequest = createDeploymentOperation(content,
+                    groupDeploymentAddresses.toArray(new ModelNode[groupDeploymentAddresses.size()]));
```
Ladicek: I wonder if `Operations.batch` would work here. Maybe not, in which case we should figure out whether the usecases in this command are rare or common, and possibly expand the `Operations` API a bit.

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
```diff
@@ -0,0 +1,211 @@
...
+
+        private List<String> serverGroups = null;
+
+        public Data(InputStream deploymentInputStream, String deploymentName, boolean autoCloseInputStream) {
+            this.deploymentInputStream = deploymentInputStream;
+            this.deploymentName = deploymentName;
+            this.autoCloseInputStream = autoCloseInputStream;
+        }
+
+        public Data(File deploymentFile) {
```
Ladicek: The constructors shouldn't be `public`.

Ladicek: In fact, probably nothing in this class should be `public`.

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
```diff
@@ -0,0 +1,211 @@
...
+     * Builder for creating command for deploying deployment to the server.
+     */
+    public static final class DeployCmdBuilder {
+        private final Data deployCmdData;
+
+        public DeployCmdBuilder(final File deploymentFile) {
+            this.deployCmdData = new Data(deploymentFile);
+        }
+
+        public DeployCmdBuilder(final InputStream deploymentInputStream, final String deploymentName,
```
Ladicek: We typically don't mark parameters `final`, unless they have to be.

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
```diff
@@ -0,0 +1,211 @@
...
+        private final Data deployCmdData;
+
+        private ServerGroupOptions(Data deployCmdData) {
+            this.deployCmdData = deployCmdData;
+        }
+
+        /**
+         * Adds list of server groups for which the deployment should be applied.
+         */
+        public Deploy toParticularServerGroups(String... serverGroups) {
```
Ladicek: I think `toServerGroups` would be shorter and better (more symmetric with `toAllServerGroups`).

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployUndeployCmdTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * This class tests deploy and undeploy commands in standalone mode.
+ */
+@Category(WildFlyTests.class)
```
Ladicek: Is there a particular reason why this test doesn't work with AS7/EAP6? I think we certainly want these commands to work there.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployUndeployCmdTest.java
```diff
@@ -0,0 +1,118 @@
...
+@RunWith(Arquillian.class)
+@RunAsClient
+public class DeployUndeployCmdTest {
+
+    private static final String DEPLOYMENT_NAME = "test-deployment.war";
+    private static final Address TEST_DEPLOYMENT_ADDRESS = Address.of("deployment", DEPLOYMENT_NAME);
```
Ladicek: I think we should add `Address.deployment`.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployUndeployCmdTest.java
```diff
@@ -0,0 +1,118 @@
...
+        ops.removeIfExists(TEST_DEPLOYMENT_ADDRESS);
+    }
+
+    private void assertDeploymentExists(String deploymentName, boolean shouldExist)
+            throws IOException, OperationException {
+        Assert.assertEquals("Deployment exists", shouldExist, ops.exists(Address.of("deployment", deploymentName)));
+    }
+
+    private void deployAsFile(Archive archive) throws CommandFailedException, IOException, OperationException {
+        File testDeploymentFile = new File(tmp.getRoot(), archive.getName());
```
Ladicek: `tmp.newFile(archive.getName())`

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployUndeployCmdTest.java
```diff
@@ -0,0 +1,118 @@
...
+    public void undeployKeepContentTest() throws OperationException, IOException, CommandFailedException {
+        deployAsFile(TEST_DEPLOYMENT);
+        client.apply(new Undeploy.Builder(TEST_DEPLOYMENT.getName()).keepContent().build());
+        assertDeploymentExists(TEST_DEPLOYMENT.getName(), true);
+        ops.removeIfExists(TEST_DEPLOYMENT_ADDRESS);
+    }
+
+    private void assertDeploymentExists(String deploymentName, boolean shouldExist)
+            throws IOException, OperationException {
+        Assert.assertEquals("Deployment exists", shouldExist, ops.exists(Address.of("deployment", deploymentName)));
```
Ladicek: The assert message only makes sense for `shouldExist == true`.

Ladicek: I mean:

```java
String assertMessage = shouldExist ? "Deployment should exist" : "Deployment shouldn't exist";
```

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Undeploy.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public void apply(OnlineCommandContext ctx) throws Exception {
+        StringBuilder cmd = new StringBuilder("undeploy ").append(deploymentName);
+        if (keepContent) {
+            cmd.append(" --keep-content");
+        }
+        if (ctx.options.isDomain) {
+            cmd.append(" --all-relevant-server-groups");
+        }
+        ctx.client.executeCli(cmd.toString());
```
Ladicek: Is there a reason you use the "local" CLI operation `undeploy` here, but for `deploy`, you assemble the `ModelNode` by hand?

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
rhatlapa: I just wanted to have also check at compile time, but I agree it is not so important in this case and having written in javadoc which options are only for domain should be satisfactory.

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
rhatlapa: I don't believe it is possible to use `Operations.batch` here. Partial help would be to expose `batch.toModelNode`. As I would be able to create the needed `ModelNode` for constructing `org.jboss.as.controller.client.Operation` which is needed here.

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
rhatlapa: You are right, good point, as I will redo it to standard builder the whole Data class should disappear.

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
rhatlapa: good point, will change the name

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployUndeployCmdTest.java
rhatlapa: I just stupidly copied the annotation without really thinking about it. It should work also with AS7/EAP6.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployUndeployCmdTest.java
rhatlapa: I believe that the message would look like: "Deployments exists, expected ...., actual ....". Maybe I've got confused by how the messages are created with TestNG vs JUnit. If it would be as I expect, I believe my original message should be fine. Will check it.

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Undeploy.java
rhatlapa: The reason is that the "local" CLI operation handles finding server groups with which the deployment is associated and I haven't found any helper method in the Helper classes related to `ModelControllerClient` to help me find them => I would have to duplicate code used by the CLI.

`deploy` directly via ModelNode allows to do the deployment just via in memory storing of the deployment, via CLI I would have to save the deployment as file.

rhatlapa:
BTW: there is also option to rewrite the `deploy` command and make use of `DeploymentPlan` class similarly as arquillian is handling deploying.

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
Ladicek: Aha, the `Operation` thing. So far, I'm only aware of a single thing that needs to attach streams to operations: deployments. Let's stick with `ModelNode`s here for now.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployUndeployCmdTest.java
Ladicek: Good! We want this command tested for all server versions for sure.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployUndeployCmdTest.java
Ladicek: For Creaper, I prefer a "should" style of assertion messages. If you expect the deployment to exist, the assertion message would read "deployment should exist"; if you expect the deployment to not exist, the assertion message would read "deployment shouldn't exist".

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Undeploy.java
Ladicek: OK then.

Ladicek:
If you feel like using `DeploymentPlan` would make the implementation easier, go for it :-) I'm fine with it both ways.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/deployments/DeployUndeployCmdTest.java
rhatlapa: that would actually create the new file, I don't want that. The export in next step shall create it.

rhatlapa:
I have incorporated your suggestions, should be ready for another round of review :-)

Ladicek:
LGTM. Please squash.

I'd also like @simkam and @ochaloup to check if the API exposes all the features they need.

Ladicek:
Also please rebase, I just pushed #61 :-) Shouldn't affect this PR, you're not doing anything version-specific.

I'd also like to do a release today, so if you don't rebase & squash soon, this will have to wait for another release. Waiting for review from @simkam and @ochaloup is not a prerequisite for merge, though.

ochaloup:
For me it's find. I haven't found anything special that I would need for more from the API.

commands/src/main/java/org/wildfly/extras/creaper/commands/deployments/Deploy.java
```diff
@@ -0,0 +1,176 @@
...
+         * </p>
+         */
+        public Deploy toServerGroups(String... serverGroups) {
+            if (this.serverGroups == null && serverGroups != null) {
+                this.serverGroups = new ArrayList<String>();
+            }
+            if (serverGroups != null) {
+                this.serverGroups.addAll(Arrays.asList(serverGroups));
+            }
+            return build();
```
simkam: I talked to Radim and he mentioned that he is not sure if this ok. I agree that this could problem in future if we eventually add more options to builder. @Ladicek wdyt?

Ladicek: It's also a bit unusual. I agree it would be better if these methods just returned `Builder` and the user always had to call `build()`.

Ladicek: On the other hand, as Radim just told me, it makes sure that the user can't call both `toServerGroups` and `toAllServerGroups`, which is nice. IMHO it would be enough to throw an exception when it's detected that the user called both of these methods, the question is how to detect this -- that would require slightly more complex encoding in the `Builder`.

rhatlapa:
I have done force update, where I have squashed the commits, rebased and modified the builder to return the builder object for `toServerGroups` and `toAllServerGroups` and emphasize in javadoc that only one of those should be used, not both of them.

Ladicek:
OK, thanks.

PR #58, opened 2016-02-01
offline version discovery

### LINKED ISSUE #7: implement version discovery in the offline case
The current version discovery is completely missing for the offline case. This must be supported for the 1.0 release.

Ladicek:
Also contains two other small commits that are not worth their own PR.

Fixes #7.


Ladicek:
Will be superseded by a PR that also incorporates a fix for #59.

PR #62, opened 2016-02-10
Added commands for add/remove security realm and ldap connection

olukas:
New tests locally pass with all profiles (wildfly10, wildfly9, wildfly8, as7).

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/security/realms/AddLdapConnection.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/security/realms/AddLdapConnection.java
@@ -1,0 +1,210 @@
+package org.wildfly.extras.creaper.commands.security.realms;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import org.wildfly.extras.creaper.commands.foundation.offline.xml.GroovyXmlTransform;
+import org.wildfly.extras.creaper.commands.foundation.offline.xml.Subtree;
+import org.wildfly.extras.creaper.core.offline.OfflineCommand;
+import org.wildfly.extras.creaper.core.offline.OfflineCommandContext;
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Batch;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+import org.wildfly.extras.creaper.core.online.operations.admin.Administration;
+
+/**
+ * Add a new LDAP outbound connection.
+ */
+public final class AddLdapConnection implements OnlineCommand, OfflineCommand {
+
+    private final String connectionName;
+    private final List<String> handlesReferralsFor;
+    private final String initialContextFactory;
+    private final String referrals;
+    private final String searchCredential;
+    private final String searchDn;
+    private final String securityRealm;
+    private final String url;
+    private final boolean replaceExisting;
+    private final Map<String, String> properties;
+
+    private AddLdapConnection(Builder builder) {
+        this.connectionName = builder.connectionName;
+        this.handlesReferralsFor = builder.handlesReferralsFor;
+        this.initialContextFactory = builder.initialContextFactory;
+        this.referrals = builder.referrals;
+        this.searchCredential = builder.searchCredential;
+        this.searchDn = builder.searchDn;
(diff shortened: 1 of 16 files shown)


commands/src/main/java/org/wildfly/extras/creaper/commands/security/realms/AddLdapConnection.java
```diff
@@ -0,0 +1,212 @@
...
+                .andOptional("url", url)
+                .andListOptional(String.class, "handles-referrals-for", resolvedHandlesReferralsFor));
+
+        if (properties != null && !properties.isEmpty()) {
+            for (Map.Entry<String, String> entry : properties.entrySet()) {
+                ops.add(ldapConnection.and("property", entry.getKey()), Values.empty()
+                        .and("value", entry.getValue())
+                );
+            }
+        }
```
Ladicek: This could possibly be a `Batch` (and `ops.batch`). Not required, but some commands to that and the error behavior is a bit nicer.

commands/src/main/java/org/wildfly/extras/creaper/commands/security/realms/AddLdapConnection.java
```diff
@@ -0,0 +1,212 @@
...
+        groovyAttributes.put("atrReferrals", referrals);
+        groovyAttributes.put("atrSearchCredential", searchCredential);
+        groovyAttributes.put("atrSearchDn", searchDn);
+        groovyAttributes.put("atrSecurityRealm", securityRealm);
+        groovyAttributes.put("atrUrl", url);
+        groovyAttributes.put("atrReplaceExisting", replaceExisting);
+        groovyAttributes.put("atrProperties", properties);
+
+        ctx.client.apply(GroovyXmlTransform.of(AddLdapConnection.class)
+                .parameters(groovyAttributes)
```
Ladicek: Not sure why you don't use `parameter` here, like this:

```java
ctx.client.apply(GroovyXmlTransform.of(AddLdapConnection.class)
        .parameter("atrConnectionName", connectionName)
        .parameter("atrHandlesReferralsFor", handlesReferralsFor)
```

Not that building a `Map` upfront is bad, but for this straightforward case, I think it doesn't make much sense.

Ladicek:
Looks good. The comments are both optional, I had to try hard to find _something_ :-)

Ladicek:
I wonder, though -- how does `AddSecurityRealm` from this PR compare to `AddHttpsSecurityRealm` in the `...commands.undertow` package? Seems that `AddSecurityRealm` is way smaller, which is weird, because I'd expect it to be more generic and eventually supersede `AddHttpsSecurityRealm`?

commands/src/main/java/org/wildfly/extras/creaper/commands/security/realms/AddLdapConnection.java
```diff
@@ -0,0 +1,212 @@
...
+        this.searchDn = builder.searchDn;
+        this.securityRealm = builder.securityRealm;
+        this.url = builder.url;
+        this.replaceExisting = builder.replaceExisting;
+        this.properties = builder.properties;
+    }
+
+    @Override
+    public void apply(OnlineCommandContext ctx) throws CliException, CommandFailedException, IOException,
+            TimeoutException, InterruptedException {
```
Ladicek: Why do all the `apply` methods still have this crappy list of thrown exception? The `*Command` interface looked like this a long time ago, but now the interface has `throws Exception` and in any case, you should only declare those exceptions that the `apply` method can actually throw.

olukas:
I will fix issues mentioned in comments. Most of them are caused by moving some parts of commands from one of our testsuite (which was developed during Creaper version ~0.3).

About relation of `AddSecurityRealm` and `AddHttpsSecurityRealm`. My idea is to develop some "simple" commands (like `AddSecurityRealm`, `AddSecurityRealmJaasAuthentication` ...) which can serve as building blocks for "complex" commands (like `AddHttpsSecurityRealm`). Then users can use more complex commands for common scenarios or prepare some special security realm which they need (e.g. something like kerberos authentication with fallback to properties and following ldap authorization).

I think using generics for complex security realm command is not good way - each of them can be very different. I think complex commands should serve, _internally_, as conjunction of smaller commands.

However this `AddSecurityRealm` command serves for adding some security realm box which can be "fill" with authentication, authorization etc.

Ladicek:
Agree with the idea about composing bigger commands from smaller ones. If you feel like that, rewriting `AddHttpsSecurityRealm` to utilize your new commands would be a good way to prove that the idea works :-) That's of course strictly optional.

olukas:
Yes, once all needed "smaller" commands will be added then `AddHttpsSecurityRealm` can be rewritten.

olukas:
Mentioned above issues were fixed.

Ladicek:
OK, looks good. Please rebase and squash.

olukas:
Squashed and rebased.

Ladicek:
Why did you keep `Fixed issues from comment` in the commit message? :-)

olukas:
Oops, thanks :) fixed

Issue #70, opened 2016-02-19
add version checks to relevant offline commands


Ladicek:
Some offline commands are version-specific and should have version checks. This wasn't possible before, but now that offline version discovery is implemented, we should add the checks retroactively.

Ladicek:
I think I spoke with @olukas about that. There might be other people who contributed commands with missing version checks in offline command. I will need to review that.

olukas:
I will add version checkers into my security-related offline commands.

Ladicek:
Thanks :-)

PR #71, opened 2016-02-19
Added version checkers to security offline commands

olukas:


Ladicek:
Addresses (part of) #70.

LGTM. This will have to wait after 1.0.0, but thanks in advance!

Ladicek:
So if we will do 0.9.6, we might just pull this in, too :-)

PR #73, opened 2016-02-22
fix the treatment of default host in managed domain

### LINKED ISSUE #72: can't execute /core-services=management/access=authorization:... in managed domain
Creaper always prepends `/host=...` to addresses beginning with `/core-service=...` in managed domain. That, however, is wrong, there are legitimate uses of `/core-service=...` addresses in managed domain (e.g. RBAC).

Currently, if there's no default host set, attempting to execute operation on a `/core-service=...` address will end up with an exception. The solution of this problem will relax this check -- if there's no default host, just execute the operation without changes.

Thanks to @cypo721 for discovering the issue!

Ladicek:
With the offline client, the notion of default host doesn't make any sense, because the client always works with a single file.

With online client, default host is used to transform operations on addresses `/core-service=...` to `/host=.../core-service=...`, and if there is no default host to prepend, an exception is thrown. That, however, is wrong, because there are legitimate uses of operations on `/core-service=...` in managed domain, e.g. RBAC. The solution is simple -- if there's no default host defined, the operation will be executed unchanged.

Fixes #72.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 ## 0.9.6 (not yet released)
 
 - offline commands for security now check the configuration file version
+- `OfflineOptions.defaultHost` and `.forHost` make no sense because
+  an offline client always works with a single file, so they are now
+  deprecated and scheduled for removal
+- if `OnlineOptions.forHost` wasn't called and `OnlineOptions.defaultHost`
+  is therefore `null`, operations against addresses `/core-service=...`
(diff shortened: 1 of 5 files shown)


Ladicek:
CC @cypo721 @rhatlapa

rhatlapa:
The changes LGTM. As we have discussed offline of this PR, either this way or introducing additional optional option for disabling the domain adjustments seems as the only possible ways to handle this case.

Issue #87, opened 2016-04-25
ManagementClient.online should declare more checked exceptions


Ladicek:
This is so that all exceptions from `OnlineOptions.createModelControllerClient` don't have to be wrapped in `IllegalStateException`. It is source-incompatible, so it has to be done in a major release.

Ladicek:
It should probably declare `throws Exception`.

Ladicek:
The newly created SSL options (see #105) are also affected.

Ladicek:
There's also `OnlineManagementClientImpl.apply` that catches `InterruptedException` and throws a different one. This needs careful examination, if the interruption status could possibly be lost in `AutomaticErrorHandlingForCommands`.

Issue #67, opened 2016-02-17
investigate integration with WildFly Launcher


Ladicek:
https://github.com/wildfly/wildfly-core/tree/master/launcher

It would have to be a separate module (`creaper-launcher`). AFAIK, it's not possible to support AS7.

Ladicek:
If we could have a launcher that would seamlessly support both embedding WildFly and launching an external WildFly process (using WildFly Launcher) and, that would be super-awesome.

PR #105, opened 2016-05-19
add support for configuring and using SSL for OnlineManagementClient

OndrejKotek:


core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
```diff
@@ -5,6 +5,8 @@
     REMOTE("remote"),
     /** Used in WildFly. Default port 9990. */
     HTTP_REMOTING("http-remoting"),
+    /** Used in WildFly. Default port 9993. */
+    HTTPS_REMOTING("https-remoting"),
```
Ladicek: How does it look in AS7 / EAP 6?

core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
```diff
@@ -260,6 +265,16 @@ public OptionalOnlineOptions disableLocalAuth() {
         }
 
         /**
+         * <p>SSL/TLS connection settings. Optional.</p>
+         *
+         * <p>For WildFly, management protocol should be set to {@code https-remoting}.</p>
+         */
+        public OptionalOnlineOptions sslOptions(SslOptions sslOptions) {
```
Ladicek: I'd call it just `ssl`.

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -0,0 +1,232 @@
...
+        private String keyStorePassword;
+        private String keyStoreType;
+        private String keyAlias;
+        private String keyPassword;
+
+        private String trustStorePath;
+        private String trustStorePassword;
+        private String trustStoreType;
+
+        /** Changes SSL/TLS protocol used for connection. Optional. TLS is used by default. */
```
Ladicek: What are the possible values?

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -0,0 +1,232 @@
...
+        public Builder protocol(String protocol) {
+            this.protocol = protocol;
+            return this;
+        }
+
+        /** Keystore with certificate for authentication. Optional. */
```
Ladicek: What if the keystore has no password?

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -0,0 +1,232 @@
...
+
+        /** Key from keystore used for authentication. Optional. */
+        public Builder key(String keyAlias, String keyPassword) {
+            this.keyAlias = keyAlias;
+            this.keyPassword = keyPassword;
+            return this;
+        }
+
+        /** Changes expected type of keystore. Optional. JKS is expected by default. */
+        public Builder keyStoreType(String keyStoreType) {
```
Ladicek: What are the possible options? Could it be an enum?

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -0,0 +1,232 @@
...
+
+        /** Truststore with trusted certificates. Optional. */
+        public Builder trustStore(String trustStorePath, String trustStorePassword) {
+            this.trustStorePath = trustStorePath;
+            this.trustStorePassword = trustStorePassword;
+            return this;
+        }
+
+        /** Changes expected type of truststore. Optional. JKS is expected by default. */
+        public Builder trustStoreType(String trustStoreType) {
```
Ladicek: Same as above.

Ladicek:
I like this. Besides the comments, there's one other thing -- loading the keystore/truststore from a file. I think `String` is not the best type for that. In Creaper, there are typically two variants: load from a file, using a `File` object to provide a path, and load from a classpath resource, using a `Class` object as a resource loader and a `String` as a resource path. See e.g. `CliFile` or `GroovyXmlTransform`. I'd like to have these two possibilities here, too. Or, if you think that loading from classpath is not very common, at least move from `String` to `File`.

So, from

```java
new SslOptions.Builder()
    .keyStore("/path/to/keystore.jds", "my password")
```

we'd move to

```java
new SslOptions.Builder()
    .keyStore(new File("/path/to/keystore.jds"), "my password")
```

If there should be the classpath variant, it would look like

```java
new SslOptions.Builder()
    .keyStore(MyClass.class, "/path/to/keystore/on/classpath.jds", "my password")
```

which doesn't look very friendly. Maybe something like

```java
new SslOptions.Builder()
    .keyStore(MyClass.class, "/path/to/keystore/on/classpath.jds").password("my password")
```

could be used. With that, we could have an easy way to signal "no password":

```java
new SslOptions.Builder()
    .keyStore(MyClass.class, "/path/to/keystore/on/classpath.jds").noPassword()
```

Same with the `File` variant.

WDYT?

Ladicek:
BTW, it's unfortunate that you have to always rethrow exceptions as `IllegalStateException` :-( You were hit by #87 -- sorry about that, but we can only fix that in a major version, which I really don't want to release very soon.

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
OndrejKotek: There is no need to change protocol in AS7 / EAP6.

core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
OndrejKotek: OK.

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
OndrejKotek: Yes, I will create enum.

https://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#SSLContext

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
OndrejKotek: `.noKeyStorePassword()` method could be added, like you suggested below.

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
OndrejKotek: It could be an enum, I'll create one.

http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#KeyStore

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
Ladicek: Aha, OK.

Anyway, you'll have to adjust the `if (data.localDefault) { ... }` condition in `OnlineOptions` constructor.

core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
```diff
@@ -389,16 +409,16 @@ public void handle(Callback[] callbacks) throws IOException, UnsupportedCallback
             }
 
             modelControllerClient = (ModelControllerClient) createMethod.invoke(null, // static method
-                    protocolName, host, port, callbackHandler, null, connectionTimeout, saslOptions);
+                    protocolName, host, port, callbackHandler, sslContext, connectionTimeout, saslOptions);
         } catch (NoSuchMethodException e) {
-            if (protocol == ManagementProtocol.HTTP_REMOTING) {
+            if (protocol == ManagementProtocol.HTTP_REMOTING || protocol == ManagementProtocol.HTTPS_REMOTING) {
                 // user asks for WildFly, but the client library is from AS7, this can't work
                 throw new IllegalStateException("The server should be WildFly (either ManagementProtocol.HTTP_REMOTING was used or the '"
```
Ladicek: Please change `ManagementProtocol.HTTP_REMOTING` to `ManagementProtocol.HTTP(S)_REMOTING` (or something similar) in the exception message.

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -0,0 +1,232 @@
...
+
+    private static char[] toCharArray(String password) {
+        if (password != null && !password.isEmpty()) {
+            return password.toCharArray();
+        }
+
+        return null;
+    }
+
+    private static void closeQuietly(FileInputStream stream) {
```
Ladicek: BTW, you can use Guava if you wish. Creaper Core already has a dependency on it.

OndrejKotek:
What about a new checked exception? Something like `ClientConfigurationFailedException` ...

Ladicek:
A new _unchecked_ exception would work. Adding new checked exceptions isn't source-compatible :-(

OndrejKotek:
Do not merge please.

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -127,59 +206,124 @@ SSLContext createSslContext() {
     }
 
     private KeyManager[] getKeyManagers() {
-        if (this.keyStorePath == null) {
+        if (this.keyStoreSource == null && this.trustStoreType != KeyStoreType.PKCS11) {
```
Ladicek: Not sure why the condition is here, and it definitely shouldn't refer to `trustStoreType`.

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -127,59 +206,124 @@ SSLContext createSslContext() {
...
+            throw new IllegalStateException("Failed to load keystore.", ex);
         } finally {
-            closeQuietly(trustStoreStream);
+            Closeables.closeQuietly(keyStoreStream);
+        }
+
+        return keyStore;
+    }
+
+    private KeyStore createRepacementKeyStore(KeyStore keyStore, char[] password) {
```
Ladicek: Care to explain why this is necessary? Preferrably in a comment in the code.

core/src/main/java/org/wildfly/extras/creaper/core/online/KeyStoreType.java
```diff
@@ -0,0 +1,53 @@
...
+    JKS("jks"),
+    /** A domain keystore is a collection of keystores presented as a single logical keystore. */
+    DKS("dks"),
+    /** A keystore backed by a PKCS #11 token. */
+    PKCS11("pkcs11"),
+    /** The transfer syntax for personal identity information as defined in PKCS #12. */
+    PKCS12("pkcs12"),
+    /** The second version of PKCS12 type keystore. */
+    PKCS12S2("pkcs12s2"),
+    ;
```
Ladicek: If some of the enum constants are impossible to use from our code (the PKCS11 thing looks particularly suspicious), they should either be removed or commented.

Ladicek:
I'm not merging anything unless you say it's ready :-) In fact, I don't even get a notification about new commits in the PR, only about comments. So if you didn't comment, I wouldn't even look at the new commit :-)

OndrejKotek:
It could be ready to merge.

README.md
```diff
@@ -285,7 +285,9 @@ an exception.)
 
 Alternatively, you can define a system property `creaper.wildfly` (its value
 is ignored, the system property just needs to be defined). This switches
-the default protocol to `http-remoting` and the default port to `9990`.
+the default protocol to `http-remoting` and the default port to `9990`
+(in case `ssl` is used, this switches the default protocol to `https-remoting`
```
Ladicek: Instead of "in case `ssl` is used", which kinda implies that there's another system property called "ssl", I'd say "in case SSL is configured".

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -0,0 +1,384 @@
...
+        /** Changes SSL/TLS protocol used for connection. Optional. {@link SslProtocol#TLS TLS} is used by default. */
+        public Builder protocol(SslProtocol protocol) {
+            this.protocol = protocol;
+            return this;
+        }
+
+        /** Keystore with certificate for authentication, from the filesystem. Optional. */
+        public Builder keyStore(File file) {
+            if (file == null) {
+                return this;
```
Ladicek: This sounds like an error, too.

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -0,0 +1,384 @@
...
+         */
+        public Builder keyStoreType(KeyStoreType keyStoreType) {
+            this.keyStoreType = keyStoreType;
+            return this;
+        }
+
+        /** Truststore with trusted certificates, from the filesystem. Optional. */
+        public Builder trustStore(File file) {
+            if (file == null) {
+                return this;
```
Ladicek: Same as above, this looks like an error.

Ladicek: I'd combine with the following condition, like this:

```java
if (file == null || !file.exists()) { ... }
```

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -0,0 +1,384 @@
...
+            throw new IllegalStateException("Failed to load keystore.", ex);
+        } finally {
+            Closeables.closeQuietly(keyStoreStream);
+        }
+
+        return keyStore;
+    }
+
+    /** Creates keystore with single entry (according to key alias) to be accessible under given password. */
+    private KeyStore createRepacementKeyStore(KeyStore keyStore, char[] password) {
```
Ladicek: Could you please explain why this is necessary? If I understand correctly, this will just copy one key from the original keystore to a new keystore. So in essence, the new keystore is a clone of the original one, the only difference being that the new keystore only contains one key from the original. Why is that needed?

core/src/main/java/org/wildfly/extras/creaper/core/online/SslOptions.java
```diff
@@ -0,0 +1,384 @@
...
+
+            return sslContext;
+        } catch (GeneralSecurityException ex) {
+            throw new IllegalStateException("Failed to create SSLContext.", ex);
+        }
+    }
+
+    private KeyManager[] getKeyManagers() {
+        // pkcs11 providers are configured externally, no keystore source required here
+        if (this.keyStoreSource == null && this.keyStoreType != KeyStoreType.PKCS11) {
```
Ladicek: Could you please explain this condition?

I understand that when `keyStoreSource` is `null` and `keyStoreType` is something different from `PKCS11`, then keystore wasn't configured and `return null` is the right thing to do.

But what happens when `keyStoreSource` isn't `null` and, at the same time, `keyStoreType` is `PKCS11`? Is that an invalid state?

Ladicek:
Oh, one thing I completely forgot -- could you add a test? Preferably one that would generate all the necessary keystores at runtime.

OndrejKotek:
Hope I have answered the questions enough. Tests will follow.

Ladicek:
Ah, OK, thanks. I think the conditions for the PKCS#11 case a bit convoluted, but whatever. LGTM barring the test.

OndrejKotek:
Tests added.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/SslOnlineManagementClientTest.java
```diff
@@ -0,0 +1,157 @@
...
+import org.wildfly.extras.creaper.core.offline.OfflineOptions;
+import org.wildfly.extras.creaper.security.KeyPairAndCertificate;
+import org.wildfly.extras.creaper.test.ManualTests;
+import org.wildfly.extras.creaper.test.WildFlyTests;
+
+/**
+ * This test <b>needs</b> a manually-controlled Arquillian container.
+ *
+ * <p>WildFly test. For AS7, there is no way to have SSL only secured management interface and to manage the server
+ * with Arquillian.</p>
```
Ladicek: One day, I will kill Arquillian with fire!

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/SslOnlineManagementClientTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Test
+    @InSequence(Integer.MIN_VALUE)
+    public void startServer() throws IOException, CommandFailedException, GeneralSecurityException {
+        this.controller.start(ManualTests.ARQUILLIAN_CONTAINER_MGMT_PROTOCOL_REMOTE);
+    }
+
+    @Test
+    @InSequence(Integer.MAX_VALUE)
```
Ladicek: This is certainly a creative use of `Integer.MIN|MAX_VALUE`, but I'd rather see `@InSequence(1)` for the first method, `@InSequence(2)` for the second and `@InSequence(3)` for the third. And the methods should be in this order, too.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/SslOnlineManagementClientTest.java
```diff
@@ -0,0 +1,157 @@
...
+                .keyStorePassword(CLIENT_KEY_PASSWORD)
+                .key(CLIENT_KEY_ALIAS, CLIENT_KEY_PASSWORD)
+                .trustStore(clientTrustStoreFile)
+                .trustStorePassword(SERVER_CERTIFICATE_PASSWORD)
+                .build();
+    }
+
+    private static void configureServerManagementInterfaces() throws CommandFailedException {
+        AddHttpsManagementSecurityRealm addHttpsSecurityRealm = new AddHttpsManagementSecurityRealm
+                .Builder(SECURITY_REALM)
```
Ladicek: `.Builder(SECURITY_REALM)` should be at the end of previous line. I know this looks slightly better in the chain, but having the constructor invocation expression split into two lines is just weird.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/SslOnlineManagementClientTest.java
```diff
@@ -0,0 +1,157 @@
...
+                .truststoreProvider(KeyStoreType.DEFAULT_TYPE.typeName().toUpperCase())
+                .build();
+
+        AddNativeManagementInterface addNativeInterface = new AddNativeManagementInterface
+                .Builder("ManagementRealm", "management-http").replaceExisting().build();
+
+        AddHttpManagementInterface addHttpsInterface = new AddHttpManagementInterface.Builder(SECURITY_REALM)
+                .secureSocketBinding("management-https").httpUpgradeEnabled(true).replaceExisting().build();
+
+        offlineClient.apply(addHttpsSecurityRealm, addNativeInterface, addHttpsInterface);
```
Ladicek: I have never seen anyone using this :-)

(Nothing wrong with it. No need to change. Just an observation :-) )

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/SslOnlineManagementClientTest.java
OndrejKotek: And what about inherited tests?

testsuite/standalone/src/test/resources/arquillian.xml
```diff
@@ -14,15 +14,30 @@
...
+                <property name="serverConfig">standalone.xml</property>
+                <property name="managementAddress">127.0.0.1</property>
+                <property name="managementPort">${management.port}</property>
+                <property name="managementProtocol">remote</property>
+
+                <property name="waitForPorts">${management.port}</property>
+                <property name="waitForPortsTimeoutInSeconds">8</property>
+            </configuration>
+        </container>
+    </group>
```
Ladicek: Oh Arquillian, how I hate thee! Marked #67 as important.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/SslOnlineManagementClientTest.java
Ladicek: Ah, inherited from the superclass! My bad, please ignore.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/SslOnlineManagementClientTest.java
```diff
@@ -0,0 +1,157 @@
...
+    public static void restoreConfiguration() throws CommandFailedException {
+        offlineClient.apply(CONFIGURATION_BACKUP.restore());
+    }
+
+    @AfterClass
+    public static void deleteKeyAndTrustStoreFiles() {
+        serverKeyStoreFile.delete();
+        serverTrustStoreFile.delete();
+        clientKeyStoreFile.delete();
+        clientTrustStoreFile.delete();
```
Ladicek: BTW, I think this doesn't pass FindBugs, does it? `mvn clean verify -DskipTests`

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/SslOnlineManagementClientTest.java
Ladicek: Here's a different idea. I personally prefer having only one `@Before` and one `@After` method. The superclass already has these (`connect` and `close`) -- WDYT about putting all the `@BeforeClass`, `@AfterClass`, `@InSequence(Integer.MIN_VALUE)` and `@InSequence(Integer.MAX_VALUE)` logic into these methods? Should be easier to understand, IMHO.

OndrejKotek: It would be easier to understand. But, It would be also slower -- starting/stopping server, creating/deleting keystores, ...

Is it really so hard to understand that we should make it slower?

Ladicek: Hmm, OK, I'm sold. Let's keep it as is.

I mean, the `@Test` methods `@InSequence(Integer.MIN_VALUE)` and `@InSequence(Integer.MAX_VALUE)` aren't really obvious at all, but you're right that it's probably the best thing we can do given JUnit, Arquillian and Surefire limitations.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/SslOnlineManagementClientTest.java
OndrejKotek: Why should it not pass? (It passes.)

Ladicek: Funny, I remember calls like this to not pass, because there's an unused return value. If it passes, the better :-)

OndrejKotek:
Are we ready to merge? Can I squash the commits and rebase?

Ladicek:
Oh, right! Please rebase and squash.

OndrejKotek:
Ready to merge.

Ladicek:
Merged, thank you!

PR #108, opened 2016-05-30
Add commands for AuditLogging

jtymel:


commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AbstractSyslogHandlerCommand.java
```diff
@@ -0,0 +1,113 @@
+package org.wildfly.extras.creaper.commands.auditlog;
+
+import org.wildfly.extras.creaper.core.offline.OfflineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+
+public abstract class AbstractSyslogHandlerCommand implements OnlineCommand, OfflineCommand {
```
Ladicek: Should it really be `public`?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddFileHandler.java
```diff
@@ -0,0 +1,133 @@
...
+import org.wildfly.extras.creaper.core.offline.OfflineCommand;
+import org.wildfly.extras.creaper.core.offline.OfflineCommandContext;
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+public class AddFileHandler implements OnlineCommand, OfflineCommand {
```
Ladicek: This name is too generic. E.g. in `org.wildfly.extras.creaper.commands.logging`, we have `AddPeriodicRotatingFileLogHandler`. This one should be called something like `AddAuditLogFileHandler`.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddHandler.java
```diff
@@ -0,0 +1,48 @@
...
+import org.wildfly.extras.creaper.commands.foundation.offline.xml.Subtree;
+import org.wildfly.extras.creaper.core.offline.OfflineCommand;
+import org.wildfly.extras.creaper.core.offline.OfflineCommandContext;
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+public class AddHandler implements OnlineCommand, OfflineCommand {
```
Ladicek: The name is too generic. It doesn't even really say what the command does.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddLogger.java
```diff
@@ -0,0 +1,98 @@
...
+import org.wildfly.extras.creaper.core.offline.OfflineCommandContext;
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+import org.wildfly.extras.creaper.core.online.operations.admin.Administration;
+
+public class AddLogger implements OnlineCommand, OfflineCommand {
```
Ladicek: Name is too generic. Should be something like `AddAuditLogger`?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddTlsSyslogHandler.java
```diff
@@ -0,0 +1,189 @@
...
+                .parameter("atrKeyPassword", keyPassword)
+                .parameter("atrKeystorePassword", keystorePassword)
+                .parameter("atrKeystorePath", keystorePath)
+                .parameter("atrKeystoreRelativeTo", keystoreRelativeTo)
+                .parameter("atrAuthenticationType", authenticationType == null ? null : authenticationType.value())
+                .parameter("atrReplaceExisting", replaceExisting)
+                .build());
+    }
+
+    public static enum AuthenticationType {
```
Ladicek: This looks like it should be top-level...?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddUdpSyslogHandler.java
```diff
@@ -0,0 +1,90 @@
...
+import org.wildfly.extras.creaper.core.CommandFailedException;
+import org.wildfly.extras.creaper.core.offline.OfflineCommandContext;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Batch;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+public class AddUdpSyslogHandler extends AbstractSyslogHandlerCommand {
```
Ladicek: This looks very similar to the TCP counterpart. I wonder how much could be reused?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/MessageTransferType.java
```diff
@@ -0,0 +1,17 @@
...
+    NON_TRANSPARENT_FRAMING("NON_TRANSPARENT_FRAMING"),
+    OCTET_COUNTING("OCTET_COUNTING");
+
+    private final String value;
+
+    MessageTransferType(String value) {
+        this.value = value;
+    }
+
+    public String value() {
```
Ladicek: Should be package-private?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/SyslogFacilityType.java
```diff
@@ -0,0 +1,39 @@
...
+    USER_LEVEL("USER_LEVEL"),
+    UUCP("UUCP");
+
+    private final String value;
+
+    SyslogFacilityType(String value) {
+        this.value = value;
+    }
+
+    public String value() {
```
Ladicek: Should be package-private?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/SyslogFormatType.java
```diff
@@ -0,0 +1,18 @@
...
+    RFC5424("RFC5424"),
+    RFC3164("RFC3164");
+
+    private final String value;
+
+    SyslogFormatType(String value) {
+        this.value = value;
+    }
+
+    public String value() {
```
Ladicek: Should be package-private?

Ladicek:
I like this, mostly. The command names are too generic -- I started to comment about that, but I'd have to add a comment to all commands :-) The names must be more specific. Ideally, all would contain the phrase `AuditLog`. If you feel like that, adding an `AuditLogging` class (similar to existing `Logging`) would be nice.

Good job on test coverage :-)

Ladicek:
Also, I'm not very familiar with audit logging. How does it work? That would explain the structure of the commands. So for audit logging, you first configure a formatter (do we have/need commands for that?), then a handler (which refers to a formatter) and finally a logger (which refers to a handler)? I guess that some javadocs for the commands would help. And if you add the `AuditLogging` entrypoint, that class's javadoc could quickly explain the overall architecture (similarly to how I explained it in this comment, if it's correct).

jtymel:
Thank you very much for the first review :)
I would say that the most important part of AuditLog is File-handler/Syslog-handler. It defines what, where and how will be logged.
These two handlers must refer the formatter they are using. We're fine with using the default `json-formatter` at the moment, therefore there is no command for creating/removing formatters right now. I can add them in next iteration of this PR if you would like to have these in library.
The final step of AuditLog settings is to refer the syslog/file handler to which will be auditable events (e.g. (un)deploy, change in configuration, server start/stop, ...) sent to within `<logger><handlers>` tag.

There is always only one `logger` node. It contains two crucial pieces of information:
* whether is AuditLogging enabled
* which syslog/file handlers are used as a destination of AuditLogging

The default configuration then looks following:
```xml
<audit-log>
    <formatters>
        <json-formatter name="json-formatter"/>
    </formatters>
    <handlers>
        <file-handler name="file" formatter="json-formatter" path="audit-log.log" relative-to="jboss.server.data.dir"/>
    </handlers>
    <logger log-boot="true" log-read-only="false" enabled="false">
        <handlers>
            <handler name="file"/>
        </handlers>
    </logger>
</audit-log>
```
I'll create an entrypoint class with javadoc in next commit.

Thanks once again for the review.

Ladicek:
Aha. If there's always exactly one logger, why do we have commands `AddLogger` and `RemoveLogger`? Should we instead have `Change[Audit]Logger`?

jtymel:
Oh, I'm terribly sorry. I definitely should have written **at most one** logger instead of **exactly one**. In that case it makes IMHO sense to keep both commands.

Ladicek:
OK, what happens if I try to `AddLogger` twice in a row?

jtymel:
It will fail with `CommandFailedException` (`Duplicate resource`)

Ladicek:
I wonder if it would make sense to keep `Remove[Audit]Logger` but replace `Add[Audit]Logger` with `ConfigureAuditLogger` that would change the existing one or add it if missing.

That would not be very future-proof, though, in case the ability to add more audit loggers will be added some time in the future...

In the end, I think I agree that we should probably keep `Add[Audit]Logger` and `Remove[Audit]Logger` and add a note to the `Add[Audit]Logger` javadoc that currently, there can only be one audit logger, so you most likely want to use `replaceExisting`.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AuditLogging.java
```diff
@@ -0,0 +1,132 @@
...
+
+    public static AuditLogDestinationHandler destinationHandler() {
+        return AuditLogDestinationHandler.INSTANCE;
+    }
+
+    public static AuditLogHandler auditLogHandler() {
+        return AuditLogHandler.INSTANCE;
+    }
+
+// ---
```
Ladicek: Indentation :-)

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AuditLogging.java
```diff
@@ -0,0 +1,132 @@
...
+    public static final class AuditLogFileHandler {
+
+        private static final AuditLogFileHandler INSTANCE = new AuditLogFileHandler();
+
+        private AuditLogFileHandler() {}
+
+        public AddAuditLogFileHandler.Builder add(String name, String path, String formatter) {
```
Ladicek: Why so many parameters if the builder's constructor only has one?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AuditLogging.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    public static final class AuditLogUdpSyslogHandler {
+
+        private static final AuditLogUdpSyslogHandler INSTANCE = new AuditLogUdpSyslogHandler();
+
+        private AuditLogUdpSyslogHandler() {}
+
+        public AddAuditLogUdpSyslogHandler.Builder add(String name, String formatter) {
```
Ladicek: Same as above -- if the builder's constructor only has one parameter, this method shouldn't have more.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AuditLogging.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    public static final class AuditLogTcpSyslogHandler {
+
+        private static final AuditLogTcpSyslogHandler INSTANCE = new AuditLogTcpSyslogHandler();
+
+        private AuditLogTcpSyslogHandler() {}
+
+        public AddAuditLogTcpSyslogHandler.Builder add(String name, String formatter) {
```
Ladicek: Same as above -- if the builder's constructor only has one parameter, this method shouldn't have more.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AuditLogging.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    public static final class AuditLogTlsSyslogHandler {
+
+        private static final AuditLogTlsSyslogHandler INSTANCE = new AuditLogTlsSyslogHandler();
+
+        private AuditLogTlsSyslogHandler() {}
+
+        public AddAuditLogTlsSyslogHandler.Builder add(String name, String formatter, String keystorePassword,
```
Ladicek: Same as above -- if the builder's constructor only has one parameter, this method shouldn't have more.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AbstractAuditLogSyslogReliableProtocolCommand.java
```diff
@@ -0,0 +1,105 @@
...
+import org.wildfly.extras.creaper.core.CommandFailedException;
+import org.wildfly.extras.creaper.core.offline.OfflineCommandContext;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Batch;
+import org.wildfly.extras.creaper.core.online.operations.OperationException;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+abstract class AbstractAuditLogSyslogReliableProtocolCommand extends AbstractAuditLogSyslogHandlerCommand {
```
Ladicek: So, I'm not sure about this class. I'm not generally fond of deeper inheritance hierarchies, and the subclasses seem to repeat a lot of the code in this class (the `apply` methods, for example).

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AuditLogging.java
Ladicek: Also appears on other places in this class.

The `// ---` comment is my personal habit, I don't require anyone to use it.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AbstractAuditLogSyslogReliableProtocolCommand.java
jtymel: Oooops, I forgot to remove both apply methods from this class. They definitely shouldn't be there.

Anyway, did I get your point right that you suggest creation of some util class that would cover the common functionality of _apply_ methods in _AddAuditLogTcpSyslogHandler_ and _AddAuditLogTlsSyslogHandler_ commands rather than extending _AbstractAuditLogSyslogReliableProtocolCommand_ (resulting in removal of this class)?

Ladicek: I'm not super-excited about this class, but if it makes sense, then keep it in place. I only found that there was something wrong :-) Once you fix it, we can re-evaluate.

Oh and BTW, there's `recconect` a couple of lines below -- that's wrong, should be `reconnect` :-)

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AbstractAuditLogSyslogReliableProtocolCommand.java
```diff
@@ -0,0 +1,37 @@
+package org.wildfly.extras.creaper.commands.auditlog;
+
+abstract class AbstractAuditLogSyslogReliableProtocolCommand extends AbstractAuditLogSyslogHandlerCommand {
```
Ladicek: OK, looking at the class now, I think it doesn't make a lot of sense, it's there just for 2 common fields. If you really feel like it should stay, then keep it in place, but I'd rather inline those 2 fields into those 2 subclasses to get rid of this level of inheritance.

Ladicek:
Generally, all `public` non-`abstract` classes of commands and their builders should be `final`.

Otherwise looks good. I'll have to have a look at the API in my IDE, but I feel like there are no blockers now.

Oh and BTW, if you push new commits, I don't get a notification. Please always add a comment when you want me to have a look. Just saying "PTAL" is enough.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AbstractAuditLogSyslogReliableProtocolCommand.java
jtymel: +1, I'll remove this class

jtymel:
Ok, _final_ modifiers added/removed and _AbstractAuditLogSyslogReliableProtocolCommand_ removed as well.
Could you please take a look on it?
I'm still thinking about the _apply_ methods in _AddAuditLogTcpSyslogHandler_, _AddAuditLogTlsSyslogHandler_ and _AddAuditLogUdpSyslogHandler_. There are common parts that are the same in each class. Do you have any idea how to avoid repeating such code? The only thing that comes to my mind is some util class(/method)..

Ladicek:
OK, I do see some similarities in the `apply` methods of those classes, but there are also differences and I don't think that it would add much value to try to extract the common parts out. Feel free to try creating an util class, you'll see for yourself if it makes sense or not.

Another option that comes to mind -- would it make sense to merge those 3 classes into one?

jtymel:
I don't think so. IMO [UDP](https://wildscribe.github.io/Wildfly/10.0.0.Final/core-service/management/access/audit/syslog-handler/protocol/udp/index.html) is quite different in comparison to [TLS](https://wildscribe.github.io/Wildfly/10.0.0.Final/core-service/management/access/audit/syslog-handler/protocol/tls/index.html). This would bring necessity to properly document each builder method in style _this parameter can be used only with TCP and UDP_ and user could be easily confused.
I'll give it a try to create some util class. However, I'm quite sceptical about the result (whether would be reasonable).

Ladicek:
OK, but you could have 1 command class with 3 builder classes. That way, there would be only 1 set of `apply` methods, yet creating the command instance would be the same as is.

jtymel:
I prepared a draft in which 3 syslog commands are merged into 1 file. The "old" classes remains unchanged as well as tests (so far).
The purpose of this change is to consider and decide which approach is cleaner (and select one of course :) ).

Could you please take a look?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogHandler.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public void apply(OnlineCommandContext ctx) throws Exception {
+        Operations ops = new Operations(ctx.client);
+        Address handlerAddress = Address.coreService("management")
+                .and("access", "audit")
+                .and("logger", "audit-log")
+                .and("handler", handlerName);
+
+        if (!ops.exists(handlerAddress)) {
```
Ladicek: Did we agree about this? I don't remember :-)

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogger.java
```diff
@@ -0,0 +1,103 @@
...
+            this.logBoot = logBoot;
+            return this;
+        }
+
+        public Builder logReadOnly(boolean logReadOnly) {
+            this.logReadOnly = logReadOnly;
+            return this;
+        }
+
+        public Builder replaceExisting() {
```
Ladicek: Should have a comment:

```java
/**
 * <b>Note that when enabled, this can cause server reload!</b>
 */
```

If there are other `replaceExisting` methods that can cause reload, they should have it too.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/RemoveAuditLogger.java
```diff
@@ -0,0 +1,32 @@
...
+import org.wildfly.extras.creaper.core.offline.OfflineCommandContext;
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+
+public final class RemoveAuditLogger implements OnlineCommand, OfflineCommand {
+
+    public RemoveAuditLogger() {
+    }
```
Ladicek: Ctor not necessary.

Ladicek:
OK, besides these 3 minor comments, I've got two more important:

1. I really like how the syslog commands were simplified. Now we can even merge `AbstractAuditLogSyslogHandlerCommand` into `AddAuditLogSyslogHandler` and everything syslog-related will be in a single class. Yay!
2. I know we discussed that already (probably), but I still don't quite understand `AddAuditLogHandler`. When looking at `AddAuditLogFileHandler` and `AddAuditLogSyslogHandler`, it's all clear directly from the name. But in what situation would I use `AddAuditLogHandler`?

Ladicek:
Oh and sorry for not reviewing sooner. The other PRs were easier to handle :-)

jtymel:
Ad point 2:
Let's assume we have following configuration:
```xml
<audit-log>
    <formatters>
        <json-formatter name="json-formatter"/>
    </formatters>
    <handlers>
        <file-handler name="file" formatter="json-formatter" path="audit-log.log" relative-to="jboss.server.data.dir"/>
        <file-handler name="notUsedDestination" formatter="json-formatter" path="not-used-audit-log.log" relative-to="jboss.dir"/>
    </handlers>
    <logger log-boot="true" log-read-only="false" enabled="true">
        <handlers>
            <!-- Handlers can be added here with AddAuditLogHandler command -->
            <handler name="file"/>
            <handler name="anotherFileHandler"/> <!-- causes IllegalStateException in application server -->
            <handler name="anotherSyslogHandler"/> <!-- causes IllegalStateException in application server -->
        </handlers>
    </logger>
</audit-log>
```

`AddAuditLogHandler` is necessary for correct and complete configuration of AuditLog.

E.g. in this example:
- `file` handler is correctly configured and the AuditLog output is passed to this destination
- `notUsedDestination` handler is correctly configured as well. However, AuditLog output is not passed there since it doesn't have its counterpart within `<audit-log><logger><handlers>` node (unlike `file`)
- `anotherFileHandler` causes an exception, since no such destination (= file/syslog handler) is defined within `<audit-log><handlers>`. Therefore application server is not able to start (or add handler to configuration through jboss-cli)
- `anotherSyslogHandler` - same as previous

So `AddAuditLogHandler` is definitely needed. I agree that its name should be probably changed though.

Ladicek:
OK, thanks for explanation, it's fairly clear from the XML, I should have looked myself :-)

Agree, we have to rename that command. Something like `AddAuditLogHandlerToAuditLogger`, but that's a bit too stupid :-) Looking at the commands for the `logging` subsystem, we use a completely different approach there. Both of them have advantages and disadvantages...

You've got any proposal? :-)

jtymel:
As I updated in the previous comment, `AddAuditLogDestinationHandler` came to my mind. Maybe `AddAuditLogOutputHandler` or something similar?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogHandler.java
jtymel: We didn't. I agree that the command should be consistent with the rest. Therefore it should probably have `replaceExisting` field.
On the other hand it only adds a handler with the only one attribute. So basically there are two possible circumstances:
- "new" handler is being written -> OK
- "new" handler is already present in the configuration -> we want to add _exactly the same_ thing (and it must  _always be the exactly same thing_) that already exists in the configuration

It looked to me odd to have `replaceExisting` field here when I wrote this class. Frankly, I still think it doesn't make much sense here. Anyway, did I get right that you insist on having `replaceExisting` everywhere? :-)

Ladicek: Ah, this is the `AddAuditLogHandler` class. No, I'm not insisting on changing this -- if the class has a more reasonable name, it can stay as is. Or if we decide to change the way how handlers are "enabled".

Ladicek:
Did you look at the commands for the `logging` subsystem? I wonder if that approach would make sense here, too.

jtymel:
Do you mean e.g. something like `AddSyslogHandler(To)AuditLog`? I think `AddAuditLogSyslogHandler` is much better name. IMHO the `AuditLog` should be at the beginning of command's name.
Or did I misunderstand what you tried to say with `logging` approach?

Ladicek:
No, when speaking about the `logging` commands, I'm not thinking about naming, I'm thinking about the general approach handlers are assigned to loggers.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogHandler.java
jtymel: Well it is maybe worth mentioning that there also exists an attribute `enabled` (that defaults to false) of `logger` node at this point :-)
```xml
    <logger log-boot="true" log-read-only="false" enabled="true">
```

If `enabled="false"` then AuditLog doesn't make any output no matter what you configured..

jtymel:
Oh, I'm sorry. Yes, the `logging` approach would make sense here. It's very similar and I guess it could be used here as well.

Ladicek:
OK, could you please provide a quick prototype? Then we can see how it looks like and finish it.

jtymel:
I tried to add PoC quickly. Guessing it should be rather placed in (newly created) `ChangeLogger`

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogger.java
```diff
@@ -95,6 +111,31 @@ public Builder replaceExisting() {
...
+            if (this.handlers == null) {
+                this.handlers = new ArrayList<String>();
+            }
+            this.handlers.addAll(Arrays.asList(handlers));
+            return this;
+        }
+
+        /**
+         * <b>Note that when enabled, this can cause server reload!</b>
+         */
```
Ladicek: The comment belongs to the `replaceExisting` method.

Ladicek:
Yea, totally agree -- we should have `ChangeAuditLogger` that will contain this. If you agree, feel free to implement that fully :-)

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogSyslogHandler.java
```diff
@@ -47,7 +80,17 @@ protected AddAuditLogSyslogHandler(TcpBuilder builder) {
...
+        this.appName = builder.appName;
+        this.facility = builder.facility;
+        this.formatter = builder.formatter;
+        this.maxFailureCount = builder.maxFailureCount;
+        this.maxLength = builder.maxLength;
+        this.syslogFormat = builder.syslogFormat;
+        this.truncate = builder.truncate;
+        this.replaceExisting = builder.replaceExisting;
+        this.host = builder.host;
+        this.port = builder.port;
```
Ladicek: You could also have a single constructor that would take `AbstractBuilder` and set all these common fields. It would be called at the beginning of these 3 constructors, saving a bit of code.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogger.java
Ladicek: This is still relevant.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AuditLogging.java
```diff
@@ -0,0 +1,126 @@
...
+        return AuditLogDestinationHandler.INSTANCE;
+    }
+
+    public static AuditLogger auditLogger() {
+        return AuditLogger.INSTANCE;
+    }
+
+    /**
+     * Configures the destinations where should be the auditable events sent to.
+     */
```
Ladicek: Comments like this shouldn't be on the classes, but on the methods that return the `INSTANCE`. Users won't typically care about these classes, but they will always care about the methods.

Ladicek:
Once you address the last comments, could you please squash? I'd like to do a final round of review, and it's easier like that for me :-) I don't expect any big findings, this seems very good now.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogger.java
jtymel: removed reloadIfRequired -> ignored

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogSyslogHandler.java
jtymel: I must have changed constructors to current state because otherwise I was getting: `variable XYZ might already have been assigned` or `variable XYZ might not have been assigned` respectively. I'm not sure about the code cleanliness though. Any ideas are welcome ;-)

jtymel:
PTAK

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/AddAuditLogSyslogHandler.java
Ladicek: Ah, true, the fun with definite assignment rules :-( I think 3 different constructors, each for one builder, are better, even with slight code duplication.

Ladicek:
Should the `@After` cleanup methods use `ChangeAuditLogger`? I'd expect that using `AddAuditLogger` would be relatively rare.

jtymel:
Ok, I rewrote one occurrence of `AddAuditLogged` to `ChangeAuditLogger`.

Much more importantly: I also added an option for removing handlers that IMHO might be useful. What do you think?

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/ChangeAuditLogger.java
```diff
@@ -91,30 +102,62 @@ public Builder logReadOnly(boolean logReadOnly) {
...
+            this.removeHandlers.addAll(Arrays.asList(handlers));
             return this;
         }
 
         public ChangeAuditLogger build() {
+            if (addHandlers != null && removeHandlers != null) {
+                List<String> auxList = new ArrayList<String>();
+                auxList.addAll(addHandlers);
+                auxList.removeAll(removeHandlers);
+                if (auxList.size() < addHandlers.size()) {
```
Ladicek:
```java
Set<String> intersection = new HashSet<String>(addHandlers);
intersection.retainAll(removeHandlers);
if (!intersection.isEmpty()) { ... }
```

Ladicek:
Aye, I thought that `ChangeAuditLogger.handlers` is a list of all handlers that will be there. If it wasn't, then the current split into `addHandlers` and `removeHandlers` makes sense. +1 for that.

Ladicek:
The change from `AddAuditLogger` to `ChangeAuditLogger` should be done everywhere :-) There are still 2  `@After` cleanup methods that use `AddAuditLogger`.

commands/src/main/java/org/wildfly/extras/creaper/commands/auditlog/ChangeAuditLogger.java
Ladicek: There's also `Sets.intersection` in Guava, but that requires to create _two_ sets instead of one, so I'd prefer to use what I wrote above.

jtymel:
Damn, I didn't notice usage of `AddAuditLogger` in `ChangeAuditLoggerOnlineTest` -> fixed.
However, the second usage in `RemoveAuditLoggerOnlineTest` makes sense. The test removes logger, therefore the logger should be added back afterwards.

Ladicek:
Ah, right you are! If you have no other plans, please squash.

jtymel:
squashed

Ladicek:
Please rebase. This should be the last time, promise :-)

jtymel:
aaand rebased :-)

Ladicek:
Please remind me again -- what's `RemoveAuditLogHandler` good for?

Ladicek:
The `AddAuditLogSyslogHandler.groovy` script is unused, no?

Ladicek:
Otherwise this is good to merge.

jtymel:
Thank you, `RemoveAuditLogHandler` is absolutely redundant and I removed it (as well as corresponding test and groovy script)
`AddAuditLogSyslogHandler.groovy` is used in `AddAuditLogSyslogHandler.java` :-) Anyway, I get your point, `AddAuditLogTcpSyslogHandler`, `AddAuditLogTlsSyslogHandler` and`AddAuditLogUdpSyslogHandler` are totally useless too, thanks again!

Squashed again

Ladicek:
Ah, I thought that the UDP/TCP/TLS variants are used and the generic variant is unused. The opposite variant is actually better, I like 3 removed files :-) Thanks, LGTM now.

Ladicek:
Merged, thanks for all the work :-) I know it's been a long time, but rest assured that there were PRs that took even longer to merge :-)

jtymel:
Thank you.
So I can be quite disappointed that I didn't even brake the record, can't I? :-)

Ladicek:
If you feel like breaking the record, please submit a PR that touches 100 or 200 files :-)

PR #109, opened 2016-05-31
Add http transport protocol.

istraka:


core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Controller provides execution of {@link ModelNode} or {@link Operation} over HTTP.
+ * Asynchronous execution is not provided. Controller does not inform about any progress, therefore, there is no point
+ * in using {@link OperationMessageHandler}.
+ * If an error occurs (server does not respond 401 or header does not contain WWW-Authenticate field after
+ * first request)
+ * {@link IllegalStateException} is thrown.
+ */
+public class HttpModelControllerClient implements ModelControllerClient {
```
Ladicek: Not `public`.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,145 @@
...
+            AuthCache authCache = new BasicAuthCache();
+            authCache.put(target, getDigestScheme());
+            localContext.setAuthCache(authCache);
+            CloseableHttpResponse response = httpClient.execute(target, buildRequest(modelNode), localContext);
+            ModelNode m = ModelNode.fromJSONStream(response.getEntity().getContent());
+            response.close();
+            return m;
+
+        } catch (MalformedChallengeException e) {
+            e.printStackTrace();
```
Ladicek: No.

pom.xml
```diff
@@ -97,6 +97,7 @@
         <version.org.jboss.shrinkwrap>1.2.3</version.org.jboss.shrinkwrap>
         <version.org.mockito.mockito-core>1.9.5</version.org.mockito.mockito-core>
         <version.xmlunit.xmlunit>1.6</version.xmlunit.xmlunit>
+        <version.org.appache.httpcomponents>4.5.2</version.org.appache.httpcomponents>
```
Ladicek: Bad ordering.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        return null;
+    }
+
+    @Override
+    public ModelNode execute(Operation operation) throws IOException {
+        return execute(operation.getOperation());
```
Ladicek: `Operation` can have `InputStream`s attached. There's a way to translate that to HTTP, we should figure it out. Or at least throw an exception if there are attachments.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
istraka: Should be the exception rethrown as `IllegalStateException` ?

core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
```diff
@@ -335,6 +339,13 @@ ModelControllerClient createModelControllerClient() throws IOException {
         if (wrappedModelControllerClient != null) {
             return wrappedModelControllerClient;
         }
+        if (protocol == ManagementProtocol.HTTP) {
+            if (username == null || password == null) {
+                throw new IllegalArgumentException("Both username and password must be specified if "
+                        + "ManagementProtocol.HTTP is used.");
```
Ladicek: Don't split the string.

pom.xml
Ladicek: Also, it's not "appache", it's "apache".

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
Ladicek: `IOException` would work, too.

Ladicek:
I'm not 100% sure on the `ManagementProtocol.HTTP` thing, but that can be solved later.

For now, I'd really like to have a test :-) A subclass of `OnlineManagementClientTest` should do the trick.

istraka:
Issues have been fixed.
Some bug fixes.
Test added.
Updated dependencies.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -33,87 +34,115 @@
...
-                .build();
         url = String.format("http://%s:%s/management", host, port);
     }
 
     private void initialize() {
+        httpClient = HttpClients.custom()
+                .setDefaultRequestConfig(RequestConfig
+                        .copy(RequestConfig.DEFAULT)
+                        .setConnectTimeout(1000)
+                        .setSocketTimeout(1000)
```
Ladicek: Should be set to `connectionTimeout` from `OnlineOptions`?

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -33,87 +34,115 @@
...
+     */
     @Override
     public ModelNode execute(Operation operation, OperationMessageHandler handler) throws IOException {
-        return execute(operation.getOperation());
+        return execute(operation);
     }
 
+    /**
+     * Method call normal execution ignoring {@code handler}
+     */
```
Ladicek: 1. Not sure what the text means, 2. noone will read the javadoc of this class, it's not a public API. All the restrictions must be documented on the place where the user chooses HTTP. Which currently is `ManagementProtocol.HTTP`, I think.

pom.xml
```diff
@@ -128,6 +129,18 @@
...
+                <artifactId>httpclient</artifactId>
+                <scope>provided</scope>
+                <version>${version.org.apache.httpcomponents.httpclient}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.httpcomponents</groupId>
+                <artifactId>httpcore</artifactId>
+                <version>${version.org.apache.httpcomponents.httpcore}</version>
+                <scope>provided</scope>
+            </dependency>
```
Ladicek: Isn't `httpcore` a transitive dependency of `httpclient`?

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpOnlineManagementClientTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class HttpOnlineManagementClientTest extends OnlineManagementClientTest {
+    private static final String USERNAME = "testuser";
+    private static final String PASS = "testpass";
+
+    @BeforeClass
+    public static void addUser() throws IOException, CommandFailedException {
+        OfflineManagementClient offlineClient = ManagementClient.offline(
+                OfflineOptions.standalone()
+                        .rootDirectory(new File(System.getProperty("user.dir"), "target/jboss-as"))
```
Ladicek: I think `new File("target/jboss-as")` will work exactly the same.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpOnlineManagementClientTest.java
```diff
@@ -0,0 +1,38 @@
...
+                        .configurationFile("standalone.xml")
+                        .build());
+        offlineClient.apply(PropertiesFileAuth.mgmtUsers().defineUser(USERNAME, PASS));
```
Ladicek: Should it be reverted after the test?

pom.xml
istraka: It is, bud provide scope is not transitive. [maven transitive](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html)

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpOnlineManagementClientTest.java
istraka: It should! thanks

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * Controller provides execution of {@link ModelNode} or {@link Operation} over HTTP.
+ * Asynchronous execution is not provided. Controller does not inform about any progress, therefore, there is no point
+ * in using {@link OperationMessageHandler}.
+ * If an error occurs (server does not respond 401 or header does not contain WWW-Authenticate field after
+ * first request)
+ * Execution on {@link Operation} is allowed with <b>no attachments</b>.
+ * {@link IllegalStateException} is thrown.
+ */
+class HttpModelControllerClient implements ModelControllerClient {
```
rhatlapa: I believe we should make it possible to allow using the `HttpModelControllerClient` via `Operations` class transparently whether it is the standard one or this one using HTTP API => it should implement `OnlineManagementClient`. 

Or is there any concern not allowing this?

rhatlapa: After discussion with @istraka , this is actually used only internally and it isn't exposed => there is no need to implement the `OnlineManagementClient` to make it work as there is still created `OnlineManagementClient`. => You can ignore my previous comment.

pom.xml
Ladicek: That's not a problem, though, no? We're not using the `httpcore` APIs directly, are we?

Ladicek:
FYI, you'll have to rebase and there will be a conflict, because #105 was just merged.

istraka:
Issues have been fixed.
Test for reload operation via HTTP.
Rebased

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,175 @@
...
+    public HttpModelControllerClient(String host, int port, String username, String password, int timeoutMillis) {
+        target = new HttpHost(host, port, "http");
+        credsProvider = new BasicCredentialsProvider();
+        credsProvider.setCredentials(
+                new AuthScope(target.getHostName(), target.getPort()),
+                new UsernamePasswordCredentials(username, password));
+
+        RequestConfig.Builder configBuilder = RequestConfig.copy(RequestConfig.DEFAULT);
+        if (timeoutMillis != NO_TIMEOUT) {
+            configBuilder.setConnectTimeout(timeoutMillis).setSocketTimeout(timeoutMillis);
```
Ladicek: I'd prefer having the chained invocation on another line. One line -- one method call.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,175 @@
...
+        }
+        requestConfig = configBuilder.build();
+        url = String.format("http://%s:%s/management", host, port);
+    }
+
+    private void initialize() {
+
+        httpClient = HttpClients.custom()
```
Ladicek: Please no :-)

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,175 @@
...
+    public ModelNode execute(ModelNode modelNode) throws IOException {
+        initialize();
+        ModelNode result;
+        try {
+            AuthCache authCache = new BasicAuthCache();
+            authCache.put(target, getDigestScheme());
+            localContext.setAuthCache(authCache);
+            CloseableHttpResponse response = httpClient.execute(target, buildRequest(modelNode), localContext);
+            result = ModelNode.fromJSONStream(response.getEntity().getContent());
+
```
Ladicek: Please no :-)

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public ModelNode execute(ModelNode modelNode) throws IOException {
+        initialize();
+        ModelNode result;
+        try {
+            AuthCache authCache = new BasicAuthCache();
+            authCache.put(target, getDigestScheme());
+            localContext.setAuthCache(authCache);
+            CloseableHttpResponse response = httpClient.execute(target, buildRequest(modelNode), localContext);
```
Ladicek: It's called "closeable", does it need to be closed explicitly? Or does closing the HTTP client close this response as well?

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,175 @@
...
+        CloseableHttpResponse response = defaultHttpClient.execute(post);
+        if (response.getStatusLine().getStatusCode() != HttpStatus.SC_UNAUTHORIZED) {
+            throw new IllegalStateException(String.format("Server responded with %d instead of %d after first request.",
+                    response.getStatusLine().getStatusCode(), HttpStatus.SC_UNAUTHORIZED));
+        }
+        if (!response.containsHeader("WWW-Authenticate")) {
+            throw new IllegalStateException("Missing WWW-Authenticate header in server response after first request.");
+        }
+        md5Auth.processChallenge(response.getHeaders("WWW-Authenticate")[0]);
+        defaultHttpClient.close();
```
Ladicek: This won't execute if an exception is thrown above. Should be in a `finally` block.

Ladicek:
Please also add a comment to the README that if users want to use the HTTP transport, they have to provide Apache HttpComponents Client 4 themselves.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/operations/admin/Reload500TimesHttpTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.test.SlowTests;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.concurrent.TimeoutException;
+
+@Category(SlowTests.class)
+@RunWith(Arquillian.class)
+public class Reload500TimesHttpTest extends Reload500TimesTest {
```
Ladicek: This doesn't really have to extend `Reload500TimesTest`, no? The inheritance doesn't give you anything. You could possibly refactor `Reload500TimesTest` so that the client is created in a `@Before` method. Then, it would make sense to inherit from that and only override the `@Before` method, inheriting both the `@Test` methods for free.

istraka: Yes, I have forgot to remove it.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
istraka: IMHO closing http client should close it as the javadoc states `Closes this stream and releases any system resources associated with it.`

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
istraka: Which of these styles is preffered?
`configBuilder
     .setConnectTimeout(timeoutMillis)
     .setSocketTimeout(timeoutMillis);`
or
`configBuilder.setConnectTimeout(timeoutMillis)
    .setSocketTimeout(timeoutMillis);`
?

Ladicek: Both of them are fine. I'd probably go for the first one, but if the variable name was too short, I'd probably switch to the second. I'll leave the decision up to you.

istraka: I like the first one.

istraka:
Issues have been fixed and README.md file has been updated.

core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
```diff
@@ -356,6 +356,12 @@ ModelControllerClient createModelControllerClient() throws IOException {
         if (wrappedModelControllerClient != null) {
             return wrappedModelControllerClient;
         }
+        if (protocol == ManagementProtocol.HTTP) {
+            if (username == null || password == null) {
+                throw new IllegalArgumentException("Both username and password must be specified if ManagementProtocol.HTTP is used.");
```
Ladicek: Actually, we shouldn't force this. If you remove the `security-realm` attribute from

```xml
            <http-interface security-realm="ManagementRealm" http-upgrade-enabled="true">
                <socket-binding http="management-http"/>
            </http-interface>
```

then the HTTP management interface is available without auth. It's not common, but it's possible. The exception should be thrown only when the first request is made and we find out that auth is required.

README.md
```diff
@@ -289,6 +289,10 @@ the default protocol to `http-remoting` and the default port to `9990`
 (in case SSL is configured, this switches the default protocol to
 `https-remoting` and the default port to `9993`).
 
+If you want to use HTTP transport for executing operation (`.protocol(ManagementProtocol.HTTP)`),
+you need to provide Apache HttpComponents Client 4 by yourself. Please notice that reload operation through HTTP
+transport may fail with `WildFly 10.0.0`.
```
Ladicek: It's the "reload" word that should be in `backticks`, not "WildFly 10.0.0".

Ladicek: Also, this interrupts the text flow. I'd put this at the end of this chapter, right above the "### Offline" chapter, and into a separate subchapter, "#### HTTP transport".

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -59,14 +59,15 @@ public HttpModelControllerClient(String host, int port, String username, String
 
         RequestConfig.Builder configBuilder = RequestConfig.copy(RequestConfig.DEFAULT);
         if (timeoutMillis != NO_TIMEOUT) {
-            configBuilder.setConnectTimeout(timeoutMillis).setSocketTimeout(timeoutMillis);
+            configBuilder
+                    .setConnectTimeout(timeoutMillis).set
```
Ladicek: The `.set` thing at the end of the line? :-)

core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
istraka: Thanks I haven't realized this. I will adjust model controller.

istraka:
Updated README.md
Issues fix.
Username and passoword are not required anymore.

Ladicek:
OK, one other comment: I still don't quite understand why you need to make an extra request (the `getDigestScheme` method) for auth. See how I create the HTTP client over here, that should work in Creaper too: https://github.com/wildfly/wildfly-core/blob/master/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/interfaces/HttpManagementInterface.java#L114

istraka:
Well, I read https://hc.apache.org/httpcomponents-client-4.5.x/tutorial/html/authentication.html#d5e717 and I haven't realized that client can do It all by itself. I will change it.

istraka:
Issues fix
    Request for obtaining realm is performed only once.
    Better use of HttpClient.
    Added ssl.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -1,108 +1,108 @@
...
+
+        registry = RegistryBuilder.<ConnectionSocketFactory>create()
+                .register("http", PlainConnectionSocketFactory.getSocketFactory())
+                .register("https", sslConnectionSocketFactory)
+                .build();
         requestConfig = configBuilder.build();
-        url = String.format("http://%s:%s/management", host, port);
+        try {
+            managementRealm = (username != null && password != null) ? getManagementRealm(url) : null;
+        } catch (IOException e) {
```
Ladicek: Propagating `IOException` should be fine here, no?

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -1,108 +1,108 @@
...
             configBuilder
                     .setConnectTimeout(timeoutMillis)
                     .setSocketTimeout(timeoutMillis);
         }
+        if (ssl != null) {
+            sslContext = ssl.createSslContext();
+        } else {
+            sslContext = org.apache.http.ssl.SSLContexts.createDefault();
+        }
+        sslConnectionSocketFactory = new SSLConnectionSocketFactory(sslContext, NoopHostnameVerifier.INSTANCE);
```
Ladicek: I'm not 100% sure about the `NoopHostnameVerifier`...

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -1,108 +1,108 @@
...
-            result = ModelNode.fromJSONStream(response.getEntity().getContent());
-        } catch (MalformedChallengeException e) {
-            throw new IOException(e);
-        } finally {
-            httpClient.close();
-        }
+        CloseableHttpClient client = getHttpClient();
+        CloseableHttpResponse response = client.execute(buildRequest(modelNode));
+        ModelNode result = parseResponse(response);
+        client.close();
```
Ladicek: `client.close` should be in a `finally` block, no?

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
istraka: Yes. Thanks

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -157,21 +157,53 @@ public OperationResponse executeOperation(Operation operation, OperationMessageH
...
+    protected ModelNode parseResponse(CloseableHttpResponse response) throws IOException {
+        ModelNode result;
+        String content = EntityUtils.toString(response.getEntity());
+        int status = response.getStatusLine().getStatusCode();
+        if (status == HttpStatus.SC_OK) {
+            result = ModelNode.fromJSONString(content);
+        } else {
+            result = new ModelNode();
+            result.get(Constants.OUTCOME).set(Constants.FAILED);
+            result.get(Constants.FAILURE_DESCRIPTION).set(content);
```
Ladicek: Hey, I didn't ask for copying _everything_ from the code in WildFly Core test suite :-) I think that the original, where response was parsed for status code 500, was better.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -157,21 +157,53 @@ public OperationResponse executeOperation(Operation operation, OperationMessageH
...
+        BasicCredentialsProvider credentialsProvider = new BasicCredentialsProvider();
+        if (username != null && password != null) {
+            credentialsProvider.setCredentials(
+                    new AuthScope(host, port, managementRealm, AuthSchemes.DIGEST),
+                    new UsernamePasswordCredentials(username, password));
+        }
+
+        return HttpClients.custom()
+                .setConnectionManager(new PoolingHttpClientConnectionManager(registry))
+                .setRetryHandler(new StandardHttpRequestRetryHandler(5, true))
```
Ladicek: I'm not 100% sure about retrying by default.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -183,10 +215,14 @@ protected DigestScheme getDigestScheme() throws IOException, MalformedChallengeE
...
             throw new IllegalStateException("Missing WWW-Authenticate header in server response after first request.");
         }
-        md5Auth.processChallenge(response.getHeaders("WWW-Authenticate")[0]);
-        return md5Auth;
+        for (HeaderElement el : response.getHeaders(HttpHeaders.WWW_AUTHENTICATE)[0].getElements()) {
+            if (el.getName().equals("Digest realm")) {
+                return el.getValue();
+            }
+        }
+        throw new IllegalStateException("Digest realm not found in in WWW-Authenticate header.");
```
Ladicek: `in in`

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
istraka: Yes, maybe for code 500. But this handle other codes too instead of just throwing an exception. Isn't it better for a user?

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
```diff
@@ -10,11 +10,15 @@
...
-     * first request) {@code IllegalStateException is thrown}
      * Execution on is allowed with <b>no attachments</b> (i.e. deployment operation).
      */
     HTTP("http"),
+    /**
+     * HTTPS transport is used. Default port 9993
+     * Asynchronous execution is not provided.
+     * Execution on is allowed with <b>no attachments</b> (i.e. deployment operation).
+     */
+    HTTPS("https"),
```
Ladicek: Hmm, OK, but then `HttpModelControllerClient.https` should check if `SslOptions` are not `null`, no?

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
Ladicek: Well that's a question. I honestly don't know.

I'm leaning towards throwing an exception. Returning a `ModelNode` with the `outcome` of `failed` means that the server did execute the operation, found that there's something wrong, rollbacked it, and sent a result. If there would be any other problem, the native client would throw an exception, and so should we.

istraka: Yes, make sense.

Ladicek:
Thanks for including HTTPS!

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
istraka: Thanks. I MUST write It down.

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
istraka: Wouldn't be better using some default SslContext in this situation?
https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/conn/ssl/SSLContexts.html
Default is used now, but SystemDefault would be even better if We will not force user to define ssl options which is IMHO a question.

Ladicek: Aha, OK.

I was thinking that maybe there should be no `HTTPS` nor `HTTPS_REMOTING` in `ManagementProtocol`, only `HTTP` and `HTTP_REMOTING`, and it would be switched to the `S` variant if `SslOptions` are defined, but that's probably a bit too magical.

Default `SSLContext` looks good. But please don't use `org.apache.http.conn.ssl.SSLContexts` if it's deprecated and points to a replacement: `org.apache.http.ssl.SSLContexts`. And `SSLContexts.createSystemDefault` sounds good, but maybe there would be unnecessary difference between `HTTPS` and `HTTPS_REMOTING` then. I think these two should be consistent in how they configure the `SSLContext`.

OndrejKotek: AFAIK, default port in EAP6/AS7 is 9443.

istraka: Ok. `org.apache.http.ssl.SSLContexts` is used. I have posted wrong link.

Ladicek: Oh, good point about the AS7 default port. Not sure what to do about that.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
istraka: Could it be an option in SslOptions? Something like `turnOfHostnameVerification()` ? With "valid only for http transport" in javadoc?

Ladicek: What happens if you use the single-parameter constructor, which doesn't take any `HostnameVerifier`?

istraka: `new DefaultHostnameVerifier(PublicSuffixMatcherLoader.getDefault())` is used.
https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/conn/ssl/DefaultHostnameVerifier.html

Ladicek: Is there something wrong with using that?

I mean, I agree that this should probably be configurable, but I'd like the HTTPS support in default configuration to be as reasonable as possible and also as close as possible to the native SSL protocol.

istraka: No. I agree with using DefaultHostnameVerifier. But this may cause errors if you WildFly SSL certificate is configured with some hostname and you have commands with localhost address.

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
istraka: Javadoc? I would keep 9993 as default.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
Ladicek: That would be IMHO a good behavior by default. An option to switch off hostname verification would be good, but it shouldn't be on by default.

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
OndrejKotek: What about taking `System.getProperty(CREAPER_WILDFLY)` into account?

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpsOnlineManagementClientTest.java
```diff
@@ -0,0 +1,53 @@
...
+    public static void removeUser() throws CommandFailedException {
+        offlineClient.apply(PropertiesFileAuth.mgmtUsers().undefineUser(USERNAME));
+    }
+
+    @Before
+    @Override
+    public void connect() throws IOException {
+        if (this.controller.isStarted(ManualTests.ARQUILLIAN_CONTAINER_MGMT_PROTOCOL_REMOTE)) {
+            client = ManagementClient.online(OnlineOptions.standalone().localDefault().ssl(sslOptions)
+                    .protocol(ManagementProtocol.HTTPS).build());
```
istraka: Chaining invocation will be one per line.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -1,108 +1,108 @@
...
         RequestConfig.Builder configBuilder = RequestConfig.copy(RequestConfig.DEFAULT);
         if (timeoutMillis != NO_TIMEOUT) {
             configBuilder
                     .setConnectTimeout(timeoutMillis)
                     .setSocketTimeout(timeoutMillis);
         }
+        if (ssl != null) {
+            sslContext = ssl.createSslContext();
+        } else {
+            sslContext = org.apache.http.ssl.SSLContexts.createDefault();
```
Ladicek: So, I think this really should be consistent between `HTTPS` and `HTTPS_REMOTING`. For now, I think we should keep it to `null` and let the `SSLConnectionSocketFactory` constructor throw an exception. I'd prefer touching the `HTTPS_REMOTING` code a bit later on.

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
Ladicek: I was thinking about that too, but `creaper.wildfly` is meant to be a convenience. Creaper users should be able to completely implement the handling of `creaper.wildfly` in their code, using `.protocol`.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpsOnlineManagementClientTest.java
Ladicek: +1 :-)

Ladicek:
Here's an idea: how about we abandon HTTPS in this PR and deal with it in a subsequent PR? I feel that it was almost ready for merge, except that a bunch of concerns regarding HTTPS suddenly emerged.

istraka:
Ok. I will revert.

core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
```diff
@@ -63,7 +63,8 @@ private OnlineOptions(Data data) {
             data.host = "localhost";
             if (data.protocol == ManagementProtocol.HTTP_REMOTING || data.protocol == ManagementProtocol.HTTP) {
                 data.port = 9990;
-            } else if (data.protocol == ManagementProtocol.HTTPS_REMOTING) {
+            } else if (data.protocol == ManagementProtocol.HTTPS_REMOTING
+                    || data.protocol == ManagementProtocol.HTTPS) {
```
OndrejKotek: Could we use following?

```
} else if (data.protocol == ManagementProtocol.HTTPS_REMOTING
        || (data.protocol == ManagementProtocol.HTTPS && System.getProperty(CREAPER_WILDFLY) != null)) {
    data.port = 9993;
} else if (data.protocol == ManagementProtocol.HTTPS && System.getProperty(CREAPER_WILDFLY) == null)) {
    data.port = 9443;
} else {

File: core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java [hunk 31]
```diff
@@ -0,0 +1,228 @@
...
+    private final SSLConnectionSocketFactory sslConnectionSocketFactory;
+
+    private HttpModelControllerClient(
+            String scheme, String host, int port, String username, String password, int timeoutMillis, SslOptions ssl) {
+        this.host = host;
+        this.port = port;
+        this.username = username;
+        this.password = password;
+        url = String.format("%s://%s:%s/management", scheme, host, port);
+        RequestConfig.Builder configBuilder = RequestConfig.copy(RequestConfig.DEFAULT);
```
Ladicek: Btw, `RequestConfig.copy(RequestConfig.DEFAULT)` seems to be completely the same as `RequestConfig.custom`.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,228 @@
...
+    private final SSLContext sslContext;
+    private final SSLConnectionSocketFactory sslConnectionSocketFactory;
+
+    private HttpModelControllerClient(
+            String scheme, String host, int port, String username, String password, int timeoutMillis, SslOptions ssl) {
+        this.host = host;
+        this.port = port;
+        this.username = username;
+        this.password = password;
+        url = String.format("%s://%s:%s/management", scheme, host, port);
```
Ladicek: I think string concatenation would work equally well :-)

Ladicek:
Oh, one more bigger thing: why is the HTTP client object always created and destroyed for each request? Sounds wasteful.

istraka:
Reverted.
Issues have been fixed.
HttpClient is created only once.
Request for obtaining management realm (for digest authentication) is performed only once.

core/pom.xml
```diff
@@ -32,6 +32,10 @@
             <artifactId>jboss-as-cli</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+        </dependency>
```
Ladicek: This is still wrongly ordered. `org.apache` belongs before `org.jboss`.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,197 @@
...
+        // registry of factories
+        Registry<ConnectionSocketFactory> registry = RegistryBuilder.<ConnectionSocketFactory>create()
+                .register("http", PlainConnectionSocketFactory.getSocketFactory())
+                .build();
+        requestConfig = configBuilder.build();
+        if (timeoutMillis != NO_TIMEOUT) {
+            configBuilder
+                    .setConnectTimeout(timeoutMillis)
+                    .setSocketTimeout(timeoutMillis);
+        }
```
Ladicek: This is repeated.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -0,0 +1,197 @@
...
+     * <b>Not supported!</b>
+     */
+    @Override
+    public AsyncFuture<OperationResponse> executeOperationAsync(Operation operation, OperationMessageHandler handler) {
+        throw new UnsupportedOperationException("Asynchronous execution is not supported by " + getClass().getName());
+    }
+
+    @Override
+    public void close() throws IOException {
+    }
```
Ladicek: Must close the HTTP client.

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
```diff
@@ -7,6 +7,14 @@
     HTTP_REMOTING("http-remoting"),
     /** Used in WildFly. Default port 9993. */
     HTTPS_REMOTING("https-remoting"),
+    /**
+     * HTTP transport is used. Default port 9990
+     * Asynchronous execution is not provided.
```
Ladicek: Creaper doesn't expose async operations, so this comment is pointless.

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
```diff
@@ -7,6 +7,14 @@
...
+     * If an error occurs (server does not respond 401 or header does not contain WWW-Authenticate field after
+     * first request) {@code IllegalStateException is thrown} (if username and password is provided)
+     * Execution on is allowed with <b>no attachments</b> (i.e. deployment operation).
+     */
```
Ladicek: Actually, this text as a whole is a bit hard to read. I'd like something like this better:

> With {@code HTTP}, the management client will not use the native management protocol, but will instead use the HTTP management endpoint. The {@code ModelNode}s will be serialized to JSON. Operation attachments are not supported (yet), so deploying will not work.

core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
```diff
@@ -356,6 +356,9 @@ ModelControllerClient createModelControllerClient() throws IOException {
         if (wrappedModelControllerClient != null) {
             return wrappedModelControllerClient;
         }
+        if (protocol == ManagementProtocol.HTTP) {
+            return new HttpModelControllerClient(host, port, username, password, connectionTimeout);
```
Ladicek: This doesn't call `connectAndWaitUntilServerBoots`.

testsuite/pom.xml
```diff
@@ -97,6 +97,10 @@
             <groupId>org.jboss.arquillian.junit</groupId>
             <artifactId>arquillian-junit-container</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+        </dependency>
```
Ladicek: Bad ordering.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpOnlineManagementClientTest.java
```diff
@@ -0,0 +1,57 @@
...
+                .auth(USERNAME, PASS)
+                .connectionTimeout(5000)
+                .build());
+    }
+
+    @Test
+    @Override
+    public void executeThroughCli_reload() throws IOException {
+        Assume.assumeTrue(client.version().greaterThanOrEqualTo(ServerVersion.VERSION_4_1_0)
+                || client.version().lessThan(ServerVersion.VERSION_2_0_0));
```
Ladicek: I thought the problem is only on WildFly 10.0.0.Final, so `assumeTrue(client.version() != ServerVersion.VERSION_4_0_0)` should be enough, no?

Ladicek:
OK, looks mostly good. The remaining comments should be easy to address.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpOnlineManagementClientTest.java
istraka: Yes. I have just run Reload500TimesHttpTest with WF8 and 9 and it passed.

istraka:
Issues have been fixed.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpOnlineManagementClientTest.java
```diff
@@ -50,8 +50,7 @@ public void connect() throws IOException {
     @Test
     @Override
     public void executeThroughCli_reload() throws IOException {
-        Assume.assumeTrue(client.version().greaterThanOrEqualTo(ServerVersion.VERSION_4_1_0)
-                || client.version().lessThan(ServerVersion.VERSION_2_0_0));
+        Assume.assumeTrue(client.version() != ServerVersion.VERSION_4_1_0);
```
Ladicek: Is that really 4.1.0? WildFly 10.0.0 is 4.0.0, no?

core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
```diff
@@ -356,9 +356,6 @@ ModelControllerClient createModelControllerClient() throws IOException {
         if (wrappedModelControllerClient != null) {
             return wrappedModelControllerClient;
         }
-        if (protocol == ManagementProtocol.HTTP) {
-            return new HttpModelControllerClient(host, port, username, password, connectionTimeout);
-        }
```
Ladicek: I actually liked this, the rest of this method is quite involved and the less we touch it, the better. I just wanted to call that one method :-)

core/src/main/java/org/wildfly/extras/creaper/core/online/ManagementProtocol.java
```diff
@@ -8,11 +8,11 @@
...
-     * HTTP transport is used. Default port 9990
-     * Asynchronous execution is not provided.
-     * If an error occurs (server does not respond 401 or header does not contain WWW-Authenticate field after
-     * first request) {@code IllegalStateException is thrown} (if username and password is provided)
-     * Execution on is allowed with <b>no attachments</b> (i.e. deployment operation).
+     * With {@code HTTP}, the management client will not use the native management protocol,
+     * but will instead use the HTTP management endpoint. The {@code ModelNode}s will be serialized to JSON. Operation
+     * attachments are not supported (yet), so deploying will not work. If username and password is not provided,
+     * no authentication method will be used.
+     * Default port 9990.
```
Ladicek: Yea, the last two sentences are needed. Thanks for fixing it for me.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpOnlineManagementClientTest.java
istraka: Aaah, yes!

istraka:
Done.

istraka:
Not done!

istraka:
Issues have been fixed. Last commit amended.

Ladicek:
OK, I'll check a bit more tomorrow and then merge, but LGTM right now. Thanks!

Ladicek:
OK, so here's a couple of cosmetics:

- `core/pom.xml` and `testsuite/pom.xml` have bad ordering of the HTTP client dependency
- the class `HttpModelControllerClient` should be `final`, its constructor should be package-private and all the `protected` methods should be `private`
- the `HttpModelControllerClient.getManagementRealm` method doesn't call `close` on the response -- I don't think that closing the HTTP client is enough
- the `HttpOnlineManagementClientTest` should override the `clientConfiguredForDomain` method with a HTTP-specific variant

LGTM then.

istraka:
Done.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -173,6 +173,7 @@ protected String getManagementRealm(String url) throws IOException {
         CloseableHttpResponse response = null;
         try {
             response = defaultHttpClient.execute(post);
+            response.close();
```
Ladicek: Should be in a `finally` block.

istraka: No. If execution fails there is no response to be closed.

Ladicek: Oh, right you are.

Ladicek:
LGTM. Please squash.

istraka:
Done.

Ladicek:
Merged, thanks. Feel free to create another PR with the HTTPS changes :-)

PR #115, opened 2016-06-14
Edit if statement from truststorePath to truststorePassword

### LINKED ISSUE #110: AddHttpsSecurityRealm, trustore won't be configured if truststore-path missing
TruststorePath is not required, for example, if specifying PKCS11 truststore.

Probably better than `if (truststorePath != null) {` would be `if (truststorePassword != null) {` , because truststorePassword is required.

jtymel:
Resolves #110


Ladicek:
Thanks.

PR #117, opened 2016-06-14
Add RemoveSocketBinding command

### LINKED ISSUE #111: Provide counterpart remove command to AddSocketBinding

jtymel:
Resolves #111

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/RemoveSocketBinding.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/RemoveSocketBinding.java
@@ -1,0 +1,47 @@
+package org.wildfly.extras.creaper.commands.socketbindings;
+
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+
+/**
+ * A command for removing socket binding from default socket binding group if not specified else.
+ */
+public final class RemoveSocketBinding implements OnlineCommand {
+
+    private final String socketBindingName;
+    private final String socketBindingGroup;
+
+    public RemoveSocketBinding(String socketBindingName) {
+        this(socketBindingName, null);
+    }
+
+    public RemoveSocketBinding(String socketBindingName, String socketBindingGroup) {
+        if (socketBindingName == null) {
+            throw new IllegalArgumentException("Name of the socket-binding must be specified as non null value");
+        }
+        if (socketBindingName.isEmpty()) {
+            throw new IllegalArgumentException("Name of the socket-binding must not be empty value");
+        }
+
+        this.socketBindingName = socketBindingName;
+        this.socketBindingGroup = socketBindingGroup;
+    }
+
+    @Override
+    public void apply(OnlineCommandContext ctx) throws Exception {
+        Operations ops = new Operations(ctx.client);
+
+        String bindingGroup = this.socketBindingGroup;
+        if (bindingGroup == null) {
+            bindingGroup = ctx.client.options().isDomain ? "full-sockets" : "standard-sockets";
+        }
+
+        Address socketBindingAddress = Address.of("socket-binding-group", bindingGroup)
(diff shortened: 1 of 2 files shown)


commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/RemoveSocketBinding.java
```diff
@@ -0,0 +1,48 @@
...
+        Operations ops = new Operations(ctx.client);
+
+        if (socketBindingGroup == null) {
+            socketBindingGroup = ctx.client.options().isDomain ? "full-sockets" : "standard-sockets";
+        }
+
+        Address socketBindingAddress = Address.of("socket-binding-group", socketBindingGroup)
+                .and("socket-binding", socketBindingName);
+
+        ops.removeIfExists(socketBindingAddress);
```
rhatlapa: If the intended behaviour is to remove it lazily => only if it exists, the whole command should be either documented as such or have it configurable. As from `RemoveSocketBinding` by default I would expect that it would fail if such resource doesn't exist.

PS: I personally like the behaviour of having it removed lazily still it is inconsistent with other remove commands which fail if resource doesn't exist which is unfriendly from UX

jtymel: Thanks, that makes sense.
-> Mentioned in javadoc

commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/RemoveSocketBinding.java
```diff
@@ -0,0 +1,47 @@
...
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+
+/**
+ * A command for removing socket binding from default socket binding group if not specified else.
+ * If specified socket-binding does not exist, no operation is performed.
+ */
+public final class RemoveSocketBinding implements OnlineCommand {
+
+    private final String socketBindingName;
+    private String socketBindingGroup;
```
Ladicek: This should really be `final` and there should be an extra variable in the `apply` method.

commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/RemoveSocketBinding.java
Ladicek: I'm really not sure about this. All existing `Remove*` commands in Creaper just do `ops.remove`, letting the command fail if the resource doesn't exist.

Do you have any specific reason for doing `ops.removeIfExists`?

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/socketbindings/RemoveSocketBindingTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        RemoveSocketBinding removeSocketBinding = new RemoveSocketBinding(SOCKET_BINDING_NAME, SOCKET_BINDING_GROUP);
+        client.apply(removeSocketBinding);
+        Assert.assertFalse(ops.exists(TEST_SOCKET_BINDING_ADDRESS));
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void removeSocketBinding_nullName() throws Exception {
+        new RemoveSocketBinding(null);
+        fail("Creating command with null handler name should throw an exception");
```
Ladicek: What handler is that? :-)

commands/src/main/java/org/wildfly/extras/creaper/commands/socketbindings/RemoveSocketBinding.java
jtymel: No, I don't. Ok, I'll rewrite it to be consistent with the rest of commands.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/socketbindings/RemoveSocketBindingTest.java
jtymel: Damn you copy&paste :)

Ladicek:
LGTM. Please squash and rebase.

jtymel:
Thanks, done.

Ladicek:
Thanks.

PR #119, opened 2016-06-20
Thrown exception's messages are better when some HTTP failure occures.

istraka:
@rhatlapa is this better?  Or would it be better to try to execute command if management realm is unable to be obtained? IMHO If user demanded authentication with username and password, there should be an error when there is no authentication. Just like `curl --digest -L -D - http://localhost:9990/management --header "Content-Type: application/json" -d '{"operation" : "read-children-types", "address" : []}' -u user:pass ` which will fail when authentication is off.

### DIFF
--- a/core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
@@ -181,26 +181,37 @@ final class HttpModelControllerClient implements ModelControllerClient {
 
         HttpPost post = new HttpPost(url);
         post.addHeader(HttpHeaders.CONTENT_TYPE, ContentType.APPLICATION_JSON.getMimeType());
+        // this is not necessary, but without this server produce parse error exception when authentication is disabled
+        // which may confuse user
+        post.setEntity(new StringEntity("[]"));
 
         CloseableHttpResponse response = null;
+        String content;
         try {
             response = defaultHttpClient.execute(post);
-            response.close();
+            content = EntityUtils.toString(response.getEntity());
         } finally {
+            if (response != null) {
+                response.close();
+            }
             defaultHttpClient.close();
         }
+        if (response.getStatusLine().getStatusCode() == HttpStatus.SC_INTERNAL_SERVER_ERROR) {
+            throw new IllegalStateException(String.format("Failed to obtain management realm name. Server responded %d instead of %d. Isn't server authentication turned off while username and password set? Content: %s",
+                    HttpStatus.SC_INTERNAL_SERVER_ERROR, HttpStatus.SC_UNAUTHORIZED, content));
+        }
         if (response.getStatusLine().getStatusCode() != HttpStatus.SC_UNAUTHORIZED) {
-            throw new IllegalStateException(String.format("Server responded with %d instead of %d after first request.",
-                    response.getStatusLine().getStatusCode(), HttpStatus.SC_UNAUTHORIZED));
+            throw new IllegalStateException(String.format("Failed to obtain management realm name. Server responded %d instead of %d. Content: %s",
+                    response.getStatusLine().getStatusCode(), HttpStatus.SC_UNAUTHORIZED, content));
         }
         if (!response.containsHeader(HttpHeaders.WWW_AUTHENTICATE)) {
-            throw new IllegalStateException("Missing WWW-Authenticate header in server response after first request.");
+            throw new IllegalStateException("Failed to obtain management realm name. Missing WWW-Authenticate header in server response.");
         }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -190,17 +190,21 @@ private String getManagementRealm(String url) throws IOException {
...
-            throw new IllegalStateException(String.format("Server responded with %d instead of %d after first request.",
+            throw new IllegalStateException(String.format("Authentication phase failed. Server responded %d instead of "
+                    + "%d. Isn't server authentication turned off while username and password set?",
                     response.getStatusLine().getStatusCode(), HttpStatus.SC_UNAUTHORIZED));
         }
         if (!response.containsHeader(HttpHeaders.WWW_AUTHENTICATE)) {
-            throw new IllegalStateException("Missing WWW-Authenticate header in server response after first request.");
+            throw new IllegalStateException("Authentication phase failed. Missing WWW-Authenticate header in server "
+                    + "response after first request and management realm needed for digest authentication is unable"
+                    + "to be obtained.");
```
Ladicek: Missing space between words :-)

Ladicek:
BTW, Creaper's Checkstyle configuration tolerates very long string literals, so you don't really have to split them unless you want to.

istraka:
Thanks. I have not realized that.

Ladicek:
Also, I really don't like "authentication phase". Auth is done for each request. The first request only tries to figure out the name of the management realm. I'd rather see something like "Failed to obtain management realm name".

istraka:
Information about possible authentication being turned off is valid only with 500 status code.
Issues fix.

Ladicek:
LGTM, but I didn't really request this change, so let's wait for @rhatlapa :-)

istraka:
Yes.

rhatlapa:
From my PoV it is enough to add information about failing to retrieve the realm name as part of the first check based on status code. No need to update the additional exception messages. Don't see added value there as it would already fail if realm name wasn't detected.

Also are you sure that 500 code means only this situation? Maybe it would be worth in such situation to also return response body which can explain more what happened.

Those are just minor suggestions from my side, otherwise it LGTM.

istraka:
No, but 500 code can mean this situation. Therefore there is hint about username and password only in this situation. Message does not says, that this is a problem.
Yes, adding content could make it even better, thanks!

Ladicek:
I take it this isn't final yet?

istraka:
Added content and fixed one minor bug. This should be final.

Ladicek:
Don't push the line breaks on me. I like my exception messages single-line. Instead of `%nContent:%n`, please just `_Content:_` (where the underscore means one space :-) ).

istraka:
Done.

Ladicek:
Did you literally put the underscores in there?

istraka:
I can not explain how did this happen. Commit is amended.

Ladicek:
LGTM. Is this ready for merge?

istraka:
Yes. Commits have been squashed.

Ladicek:
@rhatlapa ?

Ladicek:
Commit message: `occures` -> `occurs`

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -181,26 +181,35 @@ private String getManagementRealm(String url) throws IOException {
...
+        post.setEntity(new StringEntity("[]"));
 
         CloseableHttpResponse response = null;
+        String content;
         try {
             response = defaultHttpClient.execute(post);
+            content = EntityUtils.toString(response.getEntity());
             response.close();
         } finally {
             defaultHttpClient.close();
```
rhatlapa: Shouldn't be the `response.close()` in finally block?

rhatlapa:
Just one question (see above), otherwise it LGTM.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
Ladicek: I asked the exact same question in the original PR, where the answer clearly was "no". But there's a change to this piece of code, so maybe this should be reevaluated.

istraka: Yes, it make sense, because `EntityUtils.toString` may throw an exception.

istraka: Fixed.

Ladicek:
OK, thanks. Please squash and I'll merge.

istraka:
Squashed.

PR #122, opened 2016-06-28
Logging api enhancement to set logging category without need of replaceExisting

ochaloup:
I would like propose small `Logging` class api enhancement. For setting category level would be fine not need to call `replaceExisting()` for `add` command. When I want just set without thinking if exist or not.

Not adding any test as seems to me not necessary but if you are for to merge and you prefer to have test I will copy some add/replaceExisting one :)

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/logging/Logging.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/logging/Logging.java
@@ -44,6 +44,18 @@ public final class Logging {
         public RemoveLogger remove(String category) {
             return new RemoveLogger(category);
         }
+
+        /**
+         * <p>
+         * Setting the logger category when ensuring if exists.<br>
+         * For existing it replaces the log level, for non-existing
+         * it creates new category with provided level.
+         * <p>
+         * This method is equivalent to <code>add(String category).replaceExisting()</code>
+         */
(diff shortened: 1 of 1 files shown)


Ladicek:
I don't think this is that big of a deal, but if it bothers you enough to create a PR, I'm not gonna oppose :-)

I'd like to have a better name, though. I think that some Creaper commands use `define` to mean "ensure that it exists, no matter how" (and `undefine` for the opposite). WDYT?

A bit of javadoc saying something like "ensures that a logger for given category exists and is configured in given way; equivalent to add + replaceExisting" would be good :-)

ochaloup:
As I want to use it in this way is worthy for me to pass even this small change than each time need to think about adding some `replaceExisting` method.

I've followed you guidelines of the `Logging` class so previously I've decided not to mess the nice code with some comments :)

Your suggestions were implemented. The only think that I do not understand is if you prefer creating method `undefine` as well which will be 'a clone' of `remove` method?

Ladicek:
No, I don't want you to add `undefine`, that was just an example. BTW `undefine` would _not_ be a clone of `remove`, because `remove` would fail if the resource is missing. `undefine` is `if (exists) remove();`.

Ladicek:
Merged (and improved the javadoc slightly :-) ), thanks for the PR.

PR #124, opened 2016-06-29
Take transaction offline command not being available only for WFLY/EAP7

ochaloup:
Changing txn offline command behavior to admit run for EAP 6.x.


Ladicek:
LGTM.

Ladicek:
Merged, thanks.

PR #126, opened 2016-06-30
add Administration.shutdownGracefully

### LINKED ISSUE #125: add Administration.shutdownGracefully
Will have to throw an exception if the server doesn't support graceful shutdown.

Related to #120.

Ladicek:
Fixes #125.

### DIFF
--- a/core/src/main/java/org/wildfly/extras/creaper/core/online/Constants.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/online/Constants.java
@@ -37,6 +37,7 @@ public final class Constants extends ClientConstants {
     public static final String START = "start";
     public static final String STEPS = "steps";
     public static final String STOP = "stop";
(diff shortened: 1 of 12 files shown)


Ladicek:
I'm currently not sure if this is a proper way of supporting graceful shutdown in Creaper, so I'm going to leave the pull request open for some time for review.

CC @rhatlapa.

rhatlapa:
LGTM, maybe also some tests might be added, mainly some tests which would make sure it doesn't get broken for the older versions of the WildFly/AS7 servers (no-one will remove or forget to have proper handling for `timeout=0` (currently seems done properly, having test for it would give better guarantees :-)).

Ladicek:
Tests are definitely missing, thanks for the notice.

Ladicek:
Added a test. The new commits are supposed to be squashed.

rhatlapa:
Thanks, LGTM.

PR #127, opened 2016-06-30
add ReloadToOriginal

### LINKED ISSUE #123: Administration.reload should support setting the use-current-server-config parameter
The `:reload` management operation can take a parameter called `use-current-server-config`. If this parameter is set to `false` (the default is `true`) and `--read-only-server-config` was used when starting the server, the server reloads to the original configuration. This might be useful as a workaround for #14.

Ladicek:
Fixes #123.


Ladicek:
I'm currently not sure if this is a proper way of supporting "restore configuration to known point in the past" in Creaper, so I'm going to leave the pull request open for some time for review.

CC @mnovak1 @kwart.

PR #134, opened 2016-08-24
Commands for adding/removing infinispan caches

jmartisk:
This also adds support for Double in Values and ModelNodeResult, because it is needed for some properties of distributed caches.

Offline version of commands... Ugh, maybe later, aligator.

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AbstractAddCache.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AbstractAddCache.java
@@ -1,0 +1,96 @@
+package org.wildfly.extras.creaper.commands.infinispan.cache;
+
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+abstract class AbstractAddCache implements OnlineCommand {
+
+    private final CacheType cacheType;
+    private final String cacheContainer;
+    private final String name;
+
+    // cache properties common to all cache types
+    private final String jndiName;
+    private final String module;
+    private final String start;
+    private final Boolean statisticsEnabled;
+
+    protected final Address address;
+
+    @Override
+    public final void apply(OnlineCommandContext ctx) throws Exception {
+        Values values = Values.empty()
+                .andOptional("jndi-name", jndiName)
+                .andOptional("module", module)
+                .andOptional("start", start)
+                .andOptional("statistics-enabled", statisticsEnabled);
+        values = addValuesSpecificForCacheType(values);
+
+        Operations ops = new Operations(ctx.client);
+        ops.add(address, values);
+    }
+
+    abstract Values addValuesSpecificForCacheType(Values generalCacheValues);
+
+    AbstractAddCache(Builder builder, CacheType cacheType) {
+        if (builder.cacheContainer == null) {
+            throw new NullPointerException("Cache container is required");
+        }
+        this.cacheContainer = builder.cacheContainer;
+        this.cacheType = cacheType;
+        this.name = builder.name;
+        this.jndiName = builder.jndiName;
+        this.module = builder.module;
+        this.start = builder.start;
+        this.statisticsEnabled = builder.statisticsEnabled;
+        this.address = Address.subsystem("infinispan")
+                .and("cache-container", cacheContainer)
+                .and(cacheType.getType(), name);
+    }
+
+    public abstract static class Builder<THIS extends Builder> {
+
+        protected String name;
+        protected String cacheContainer;
+        protected String jndiName;
+        protected String module;
+        protected String start;
+        protected Boolean statisticsEnabled;
+
+        protected Builder(String name) {
+            if (name == null || name.isEmpty()) {
+                throw new IllegalArgumentException("Name of the cache must be specified");
+            }
+            this.name = name;
+        }
+
+        public final THIS cacheContainer(String cacheContainer) {
+            this.cacheContainer = cacheContainer;
+            return (THIS) this;
+        }
+
+        public final THIS jndiName(String jndiName) {
+            this.jndiName = jndiName;
+            return (THIS) this;
+        }
+
+        public final THIS module(String module) {
+            this.module = module;
+            return (THIS) this;
+        }
+
+        public final THIS start(String start) {
+            this.start = start;
+            return (THIS) this;
+        }
+
+        public final THIS statisticsEnabled(Boolean statisticsEnabled) {
+            this.statisticsEnabled = statisticsEnabled;
+            return (THIS) this;
+        }
+
+    }
+}
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AddDistributedCache.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AddDistributedCache.java
@@ -1,0 +1,120 @@
+package org.wildfly.extras.creaper.commands.infinispan.cache;
+
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+public final class AddDistributedCache extends AbstractAddCache {
+
+    private final Boolean asyncMarshalling;
+    private final CacheMode mode;
+    private final Long queueFlushInterval;
+    private final Long remoteTimeout;
+    private final Double capacityFactor;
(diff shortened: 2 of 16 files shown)


commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AbstractAddCache.java
```diff
@@ -0,0 +1,100 @@
...
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+
+/**
+ * @author Jan Martiska
+ */
+public abstract class AbstractAddCache implements OnlineCommand {
```
Ladicek: We don't do `@author`s in Creaper. It's visible from Git history.

Also, I don't think this class should be `public`.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AbstractAddCache.java
```diff
@@ -0,0 +1,100 @@
...
+        this.jndiName = builder.jndiName;
+        this.module = builder.module;
+        this.start = builder.start;
+        this.statisticsEnabled = builder.statisticsEnabled;
+        this.address = Address.subsystem("infinispan")
+                .and("cache-container", cacheContainer)
+                .and(cacheType.getType(), name);
+    }
+
+    public static class Builder<THIS extends Builder> {
```
Ladicek: This should probably be `abstract`.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AbstractAddCache.java
```diff
@@ -0,0 +1,100 @@
...
+
+        String name;
+        String cacheContainer;
+        String jndiName;
+        String module;
+        String start;
+        Boolean statisticsEnabled;
```
Ladicek: If these are supposed to be visible by subclasses, I'd make them `protected` (though it's not needed).

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AbstractAddCache.java
```diff
@@ -0,0 +1,100 @@
...
+
+        public Builder(String name) {
```
Ladicek: Should probably be `protected`.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AddDistributedCache.java
```diff
@@ -0,0 +1,115 @@
+package org.wildfly.extras.creaper.commands.infinispan.cache;
+
+import org.wildfly.extras.creaper.core.online.operations.Values;
+
+/**
+ * @author Jan Martiska
+ */
+public class AddDistributedCache extends AbstractAddCache {
```
Ladicek: `final`

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AddDistributedCache.java
```diff
@@ -0,0 +1,115 @@
...
+    private final CacheMode mode;
+    private final Long queueFlushInterval;
+    private final Long remoteTimeout;
+    private final Double capacityFactor;
+    private final ConsistentHashStrategy consistentHashStrategy;
+    private final Long l1lifespan;
+    private final Integer owners;
+    private final Integer segments;
+
+    public AddDistributedCache(Builder builder) {
```
Ladicek: `private`, should only be called from the builder.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AddDistributedCache.java
```diff
@@ -0,0 +1,115 @@
...
+                .andOptional("remote-timeout", remoteTimeout)
+                .andOptional("capacity-factor", capacityFactor)
+                .andOptional("consistent-hash-strategy",
+                        consistentHashStrategy != null ? consistentHashStrategy.getType() : null)
+                .andOptional("l1-lifespan", l1lifespan)
+                .andOptional("owners", owners)
+                .andOptional("segments", segments);
+    }
+
+    public static class Builder extends AbstractAddCache.Builder<Builder> {
```
Ladicek: `final`

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Operations;
+
+/**
+ * @author Jan Martiska
+ */
+public class RemoveCache implements OnlineCommand {
```
Ladicek: `final`

core/src/main/java/org/wildfly/extras/creaper/core/online/operations/Values.java
```diff
@@ -157,6 +161,12 @@ public Values and(String name, String value) {
         return new Values(newList);
     }
 
+    public Values and(String name, Double value) {
```
Ladicek: This should use primitive `double`, similarly to other primitive types.

core/src/main/java/org/wildfly/extras/creaper/core/online/operations/Values.java
```diff
@@ -243,9 +258,21 @@ public Values andList(String name, ModelNode... value) {
         return new Values(newList);
     }
 
+    public Values andList(String name, Double... value) {
```
Ladicek: This should use primitive `double`, similarly to other primitive types.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AddDistributedCache.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    final Values addValuesSpecificForCacheType(Values generalCacheValues) {
+        return generalCacheValues
+                .andOptional("async-marshalling", asyncMarshalling)
+                .andOptional("mode", mode != null ? mode.getMode() : null)
+                .andOptional("queue-flush-interval", queueFlushInterval)
+                .andOptional("remote-timeout", remoteTimeout)
+                .andOptional("capacity-factor", capacityFactor)
+                .andOptional("consistent-hash-strategy",
+                        consistentHashStrategy != null ? consistentHashStrategy.getType() : null)
```
Ladicek: The test says that `capacity-factor` and `consistent-hash-strategy` are WFLY 9+ only -- if so, we should throw an exception if they are set and current server is < WFLY 9. Or silently ignore them, but I think that we typically do the exception thing.

Ladicek:
All comments for `AddDistributedCache` also apply to other `Add*Cache` commands.

Generally, nice work, the comments are all minor, I don't think this will require any additional substantial work, just stylistics.

Ladicek:
BTW, offline variants are _not_ required.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/AddDistributedCache.java
jmartisk: Personally, throwing an exception doesn't seem quite right to me. The test would have to contain `if`s to decide whether to add the params or not. And wildfly doesn't complain either, when you pass a parameter which doesn't do anything. Wouldn't a WARN log do a better job?

Ladicek: I'm fine with skipping those parameters, if the builder methods have a javadoc stating something like "{@code capacity-factor} is only supported on WildFly 9 and above, this setting will be ignored on previous versions".

jmartisk: Ok, that sounds reasonable.

jmartisk:
I did the changes as you requested, I hope I didn't forget anything.
The comments about unsupported params currently only apply to distributed caches, I didn't find any other parameters which aren't supported (tried all the way back to EAP 6.4).

Ladicek:
I've made a couple of stylistic changes myself and pushed. Thanks!

PR #135, opened 2016-08-24
Add Travis CI configuration for creaper

kwart:
This also requires enabling access to the repository in the Travis CI. 

https://travis-ci.org/profile/wildfly-extras

@Ladicek Could you enable it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: java
+
+jdk:
+  - oraclejdk8
+
+script: "mvn verify"


Ladicek:
What is it going to do? I'd like `mvn clean verify`.

kwart:
https://docs.travis-ci.com/user/languages/java

We can specify steps, by default it does
```
mvn test
```

and before running the build, Travis CI installs dependencies:
```
mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```

kwart:
Are there any integration tests, so the `verify` is required? Or do we have just unit tests and the `test` phase is enough?

Ladicek:
OK, my Travis skill is zero, could you please make it do `verify`? Static checks are done in that phase (CheckStyle, Findbugs, CodeNarc).

kwart:
Updated.

Ladicek:
OK, thanks. Looks reasonable, not sure how to test it other than merging this :-)

PR #136, opened 2016-08-24
Add maven settings.xml with JBoss public repository configured

kwart:


CONTRIBUTING.md
```diff
@@ -27,6 +27,9 @@ significantly, this can of course be reconsidered.
 Creaper uses Maven for build. Some build dependencies are only present in
 the JBoss.org Maven repository, so it's recommended to
 [configure Maven to use the JBoss Repository](https://developer.jboss.org/wiki/MavenGettingStarted-Developers).
+You can use custom `settings.xml` where the repository is configured already.
+The file is located in the project root folder and can be activated by
```
Ladicek: Folder is a Windows terminology. Please use "directory" :-)

CONTRIBUTING.md
```diff
@@ -27,6 +27,9 @@ significantly, this can of course be reconsidered.
...
+providing `-s` maven parameter (`mvn -s settings.xml ...`).
```
Ladicek: Maven

kwart:
Setting the `before_script` is not sufficient. Let's try sth. else. I'll update the commit with suggested fixes.

Ladicek:
I'd like to get rid of `before_script` entirely. I'm not sure why it's required -- at least for Creaper, if the CI only did `mvn clean verify -s settings.xml`, that would be entirely sufficient.

Ladicek:
Would `before_script: echo LOL` work? :-)

kwart:
The correct phase is not `before_script` but `install`. And IMO the `echo LOL` should work. I'll change it.

Ladicek:
I have no idea why the tests failed, but I found that we should probably do this instead of `echo LOL`: https://docs.travis-ci.com/user/customizing-the-build/#Skipping-the-Installation-Step

kwart:
I like more the `echo LOL` than the boring `true`.

kwart:
Trying `STANDARD` [ci evironment](https://docs.travis-ci.com/user/ci-environment/) as next step.

Ladicek:
The standard env has IPv6 enabled, according to the docs, which might affect things a bit / a lot :-)

Ladicek:
Passed, cool! BTW, what's `-Dmaven.test.redirectTestOutputToFile=true`? I guess it pollutes the console less, which is good, but is the file accessible somewhere in case something fails?

kwart:
Yes, the [description](http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#redirectTestOutputToFile) says: Set this to "true" to redirect the unit test standard output to a file (found in reportsDirectory/testName-output.txt).

kwart:
I'm thinking about few more `mvn` options `"-B -fae -q"` (batch, fail-at-end, quiet). WDYT?

Ladicek:
I'm not sure what `-B` would bring us, but it can't hurt either. `-fae` sounds like a good idea.

`-q` -- what does it do? I'd like to have as much output available as possible in case of failures. I guess there's no chance I could possibly access the build workspace with Travis, so I'm not exactly a fan of reducing the console output.

kwart:
The quiet output only show errors. But as far as we are under 4MB console output size, we are fine without this switch. So just adding `-B` and `-fae`.

Ladicek:
Merged, thanks. I was thinking about doing this myself, but was too lazy :-)

PR #145, opened 2016-11-14
Configuration backup/restore in online mode

mchoma:
The ConfigurationFileBackup set of offline commands facilitate taking a backup of the server configuration and later restoring it. It would be very useful to have the same possibility in the online mode, e.g. using the existing snapshot function.

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/foundation/online/SnapshotBackup.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/foundation/online/SnapshotBackup.java
@@ -1,0 +1,122 @@
+package org.wildfly.extras.creaper.commands.foundation.online;
+
+import java.io.IOException;
+import java.util.concurrent.TimeoutException;
+
+import org.jboss.logging.Logger;
+import org.wildfly.extras.creaper.core.CommandFailedException;
+import org.wildfly.extras.creaper.core.online.CliException;
+import org.wildfly.extras.creaper.core.online.ModelNodeResult;
+import org.wildfly.extras.creaper.core.online.OnlineCommand;
+import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
+import org.wildfly.extras.creaper.core.online.operations.admin.ReloadToSnapshot;
+
+import com.google.common.base.Strings;
+
+/**
+ * Provides a pair of online commands to backup and then restore application
+ * server configuration while the server is running. Backup consists of taking a
+ * snapshot, restore then means reloading the server from the snapshot.
+ *
+ * <p>
+ * The {@code backup} command must be applied before {@code restore},
+ * {@code backup} can be applied once, and then {@code restore} can be applied
+ * many times. If any one of these rules is violated, an exception is thrown.
+ * For special circumstances, when the backup that was already acquired is no
+ * longer needed and is not going to be restored, a {@code destroy} command is
+ * provided. If there was no backup acquired, the destroy command does nothing.
+ * </p>
+ */
+public final class SnapshotBackup {
+    private static final Logger log = Logger.getLogger(SnapshotBackup.class);
+
+    private String snapshotName; // null <=> backup wasn't acquired, can't restore
+
+    private final OnlineCommand backupPart = new OnlineCommand() {
+        @Override
+        public void apply(OnlineCommandContext occ) throws CommandFailedException, IOException, CliException {
+            if (SnapshotBackup.this.snapshotName != null) {
+                throw new CommandFailedException("Snapshot was already taken: "
+                        + SnapshotBackup.this.snapshotName);
+            }
+            ModelNodeResult modelNodeResult = occ.client.execute(":take-snapshot");
+            // Ensure operation succeeded and value is returned
+            modelNodeResult.assertDefinedValue("Taking snapshot failed.");
+            String snaphotAbsolutePath = modelNodeResult.stringValue();
+            SnapshotBackup.this.snapshotName = getSnapshotName(snaphotAbsolutePath);
+        }
+
+        /**
+         * Extract file name of absolute path to snapshot
+         *
+         * <p>
+         * Also handles situation when server and client are running on different platforms.
+         * </p>
+         *
+         * @param snapshotAbsolutePath Absolute path to snapshot
+         * @return name of snapshot or empty string if snapshotAbsolutePath is null or empty.
+         */
+        private String getSnapshotName(String snapshotAbsolutePath) {
+            if (Strings.isNullOrEmpty(snapshotAbsolutePath)) {
+                return "";
+            }
+            // Windows file names can't contain "/", so if there's a "/" in the full path, it's not Windows.
+            String fileSeparator = "/";
+            if (!snapshotAbsolutePath.contains("/")) {
+                // absolute path is from windows environment
+                fileSeparator = "\\";
+            }
+            return snapshotAbsolutePath.substring(snapshotAbsolutePath.lastIndexOf(fileSeparator) + 1);
+        }
+
+        @Override
+        public String toString() {
+            return "SnapshotBackup.backup";
+        }
+    };
+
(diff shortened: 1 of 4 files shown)


commands/src/main/java/org/wildfly/extras/creaper/commands/foundation/online/SnapshotBackup.java
```diff
@@ -0,0 +1,104 @@
...
+                throw new CommandFailedException("Snapshot was already backed up as "
+                        + SnapshotBackup.this.snapshotName);
+            }
+            String snapshot = "";
+            ModelNodeResult modelNodeResult = occ.client.execute(":take-snapshot");
+            if (modelNodeResult.isSuccess() && modelNodeResult.hasDefinedValue()) {
+                snapshot = modelNodeResult.stringValue();
+            }
+            // returns absolute path, but we need just the name of snapshot
+            snapshot = snapshot.substring(snapshot.lastIndexOf(File.separator) + 1);
```
mchoma: Any idea how to handle this properly? Thing is snapshotName as value of `--server-config` parameter for `reload` could not be absolute path. 
Note, `File.separator` is already used in AddModule.

testsuite/pom.xml
```diff
@@ -28,12 +28,16 @@
         <version.wildfly10>10.0.0.Final</version.wildfly10>
         <version.wildfly10.core>2.0.10.Final</version.wildfly10.core>
         <version.wildfly10.arquillian>1.1.0.Alpha1</version.wildfly10.arquillian>
+        <version.wildfly11>11.0.0.Alpha1-SNAPSHOT</version.wildfly11>
```
mchoma: It can be replaced with `<version.wildfly11>11.0.0.Alpha1</version.wildfly11>`.
It will be first valid public value. It can lead user to realize what is going on here. Better than SNAPSHOT, which looks like a bug.

commands/src/main/java/org/wildfly/extras/creaper/commands/foundation/online/SnapshotBackup.java
Ladicek: Well file names on Windows can contain neither `/` nor `\`. So if the string contains `/`, it obviously isn't Windows and you can split on `/`, otherwise you split on `\`.

Re `AddModule`: this command only makes sense when running on the same system as the server. That doesn't apply in this situation.

testsuite/pom.xml
Ladicek: Hmm, OK, but at least add a comment. Something like:

```xml
<!-- doesn't exist yet; update when there's a public release of WildFly 11 -->
<version.wildfly11>11.0.0.Alpha1</version.wildfly11>
```

And a similar comment to the `wildfly11` profile.

commands/src/main/java/org/wildfly/extras/creaper/commands/foundation/online/SnapshotBackup.java
```diff
@@ -35,20 +36,39 @@
...
+         * Extract file name of absolute path to snapshot
+         *
+         * <p>
+         * Also handles situation when server and client are running on different platforms.
+         * </p>
+         *
+         * @param snapshotAbsolutePath Absolute path to snapshot
+         * @return name of snapshot
+         */
+        private String getSnapshotName(String snapshotAbsolutePath) {
```
Ladicek: As the first line of the method body, I'd expect a comment along the lines of `Windows file names can't contain /, so if there's a / in the full path, it's not Windows`.

Ladicek:
OK, looks mostly good. When you fix the one last thing, please also squash.

mchoma:
PR now contains just squashed commit.

commands/src/main/java/org/wildfly/extras/creaper/commands/foundation/online/SnapshotBackup.java
```diff
@@ -0,0 +1,125 @@
...
+            return "SnapshotBackup.backup";
+        }
+    };
+
+    private final OnlineCommand restorePart = new OnlineCommand() {
+        @Override
+        public void apply(OnlineCommandContext occ) throws CommandFailedException, IOException, InterruptedException,
+            TimeoutException {
+            if (SnapshotBackup.this.snapshotName == null) {
+                throw new CommandFailedException("There's no snapshot backup to restore");
```
Ladicek: Ah, one more! :-) `There's no snapshot to restore`

testsuite/pom.xml
```diff
@@ -28,12 +28,17 @@
...
         <version.wildfly10.arquillian>1.1.0.Alpha1</version.wildfly10.arquillian>
+        <!-- doesn't exist yet; update when there's a public release of WildFly 11 -->
+        <version.wildfly11>11.0.0.Alpha1</version.wildfly11>
+        <version.wildfly11.core>3.0.0.Alpha10</version.wildfly11.core>
+        <version.wildfly11.arquillian>2.0.0.Final</version.wildfly11.arquillian>
 
         <managementPort.as7>9999</managementPort.as7>
         <managementPort.wildfly>9990</managementPort.wildfly>
 
         <ignoredCategory.as7>org.wildfly.extras.creaper.test.WildFlyTests</ignoredCategory.as7>
```
Ladicek: Now I'm thinking... should this line also contain `WildFly11Tests`?

mchoma: I think no, because SnapshotBackupTest is declared as Category {WildFlyTests, WildFly11Tests}

Ladicek: So the `WildFly11Tests` category is meant as a "subset" of `WildFlyTests`? And everytime a test is in category `WildFly11Tests`, it also has to be in category `WildFlyTests`? That kinda sorta makes sense, but should be at least documented.

mchoma: I have added comment to WildFly11Tests.

commands/src/main/java/org/wildfly/extras/creaper/commands/foundation/online/SnapshotBackup.java
```diff
@@ -0,0 +1,125 @@
...
+            }
+            String snapshotAbsolutePath = "";
+            ModelNodeResult modelNodeResult = occ.client.execute(":take-snapshot");
+            if (modelNodeResult.isSuccess() && modelNodeResult.hasDefinedValue()) {
+                snapshotAbsolutePath = modelNodeResult.stringValue();
+            }
+            // returns absolute path, but we need just the name of snapshot
+            String snapshotName = getSnapshotName(snapshotAbsolutePath);
+            if (Strings.isNullOrEmpty(snapshotAbsolutePath)) {
+                throw new CommandFailedException("Unable to take snapshot");
```
Ladicek: Ah, here's an idea: if the `:take-snapshot` operation fails, we're silently dropping the error message. That's bad. If there's one, it should be added to this exception.

mchoma: Looking into the code, I believe execute() should take care of that - if result is failed throws exception. Now it seems to me this check is unnecessary. Also unnecessary is `if (modelNodeResult.isSuccess()` WDYT?

Ladicek: `OnlineManagementClient.execute` won't throw in case of an error. However, you can do `ModelNodeResult.assertSuccess()`, which would be enough in this case IMHO.

Ladicek:
Could you please remove the `[14]` from the commit message? I'll merge then.

mchoma:
Done

Ladicek:
Not really, you only renamed the PR :-)

mchoma:
Done

mchoma:
I can't believe that :O

Ladicek:
I'd like to point out that there were PRs that took more than a _month_ to merge. Compared to that, this one was pretty fast ;-)

PR #148, opened 2017-01-12
Retrieve last value in address

rhatlapa:
This is useful as address points to resource in management API tree and
the last item is the resource name, which this way can be easily
retrieved.

### DIFF
--- a/core/src/main/java/org/wildfly/extras/creaper/core/online/operations/Address.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/online/operations/Address.java
@@ -9,7 +9,8 @@ import java.util.List;
 
 /**
  * <p>An address in the management tree. It is a sequence of string pairs ({@code key=value}), possibly empty. This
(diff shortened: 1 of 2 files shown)


Ladicek:
Frankly, I didn't expect this kind of usage. The entire point of `Address` was to be opaque.

I'm not entirely opposed to adding it, but could you elaborate on your usecase?

rhatlapa:
The usecase is mainly undesired duplicity of the resource name, once for the resource name as such and once when creating `Address`.

E.g. in HAL testsuite (testsuite for web console) you often need the resource name and the `Address` as the proper propagation of the action via web console is verified via management model.

Ladicek:
Thinking about this some more, I would agree with adding a way of exposing all of the data in some structured fashion.

For example, `StringPair` could be made `public` and `Address` could be `Iterable<StringPair>`. Well not exactly this, but something along these lines :-)

So if you have some ideas about an API to expose all of the data in the address, I'm all ears. In the meantime, I'll merge this one.

PR #152, opened 2017-02-04
urn:wildfly check in SubsystemSubtreeLocator

rsvoboda:
urn:wildfly check in SubsystemSubtreeLocator, needed for new subsystems like elytron

I think https://github.com/jboss-security-qe/creaper/commit/5e02e4303982dd440ebe6ae594037a9573763916 can't be back-ported as is. It mixes commands and core infrastructure enhancements (e.g. this change).


rhatlapa:
Thanks @rsvoboda LGTM, merging.

Ladicek:
Thanks @rhatlapa for taking care of this!

PR #153, opened 2017-02-08
Log creaper commands in JSON format in debug

### LINKED ISSUE #151: Log details of creaper commands
Here is the idea. Enable TRACE logging of commands, by extending OperationsModelNodeBuilder with logging of op.toJSONString(false).
`{
    "operation" : "add",
    "address" : [
        {
            "subsystem" : "elytron"
        },
        {
            "key-store" : "CreaperTestKeyStore"
        }
    ],
    "name" : "CreaperTestKeyStore",
    "type" : "JKS",
    "credential-reference" : {"clear-text" : "password"}
}`

Sometimes it come useful to see exactly what is going on in server configuration. Also this output could be used to build reproducers. Copy output from log and pass it to 
`
curl --digest -L -D - http://localhost:9990/management -u admin:asdasd --header "Content-Type: application/json"  -d '{"operation":"reload","name":"", "address":{"host" : "master"},"json.pretty":1}' 
`
Now I only get:
`INFO: Applying command org.wildfly.extras.creaper.commands.elytron.tls.AddKeyStore@7d3d101b`

@Ladicek , WDYT?

mchoma:
Resolves #151 

It producelog output as:
INFO: Executing operation /core-service=management/security-realm=creaperSecRealm:add
JSON format:
{
    "operation" : "add",
    "address" : [
        {
            "core-service" : "management"
        },
        {
            "security-realm" : "creaperSecRealm"
        }
    ]
}

### DIFF
--- a/core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineManagementClientImpl.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineManagementClientImpl.java
@@ -149,7 +149,8 @@ final class OnlineManagementClientImpl implements OnlineManagementClient {
     public ModelNodeResult execute(ModelNode operation) throws IOException {
         checkClosed();
         operation = adjustOperationForDomain.adjust(operation);
-        log.debugf("Executing operation %s", ModelNodeOperationToCliString.convert(operation));
+        log.debugf("Executing operation %s\nJSON format:\n%s", ModelNodeOperationToCliString.convert(operation),
+                operation.toJSONString(false));
         ModelNode result = client.execute(operation);
         return new ModelNodeResult(result);
     }


core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineManagementClientImpl.java
```diff
@@ -149,7 +149,8 @@ public void apply(Iterable<OnlineCommand> commands) throws CommandFailedExceptio
     public ModelNodeResult execute(ModelNode operation) throws IOException {
         checkClosed();
         operation = adjustOperationForDomain.adjust(operation);
-        log.debugf("Executing operation %s", ModelNodeOperationToCliString.convert(operation));
+        log.debugf("Executing operation %s\nJSON format:\n%s", ModelNodeOperationToCliString.convert(operation),
+                operation.toJSONString(false));
```
rhatlapa: I believe we should be consistent with other `execute` methods, I suggest to do similar update mainly for `execute` on `Operation`. Not sure whether it would make sense to have it also for `execute` on `String` for some debugging purposes, I guess not.

Ladicek: This should absolutely be added to `execute(Operation)` too, yes. There's no way to do that for `execute(String)`, though.

One idea: I'm not terribly happy about newlines in a log message, how about adding it as a second log, possibly to the TRACE level?

mchoma: Thanks for noticing. I have overlooked `operation.getOperation()` returns `ModelNode`.

Regarding newlines. I was considering both options. I have intentionally decided for this option, because of selecting JSON string. It is ensured than no unintentional character will be copied.

Ladicek: BTW, `Operation.getOperation` gives you a `ModelNode`, but the entire purpose of the `Operation` class is to have attachments, the `ModelNode` doesn't contain everything. Just a notice.

I'm still not exactly fond of expanding a single-line log message to 3 lines, but I guess I can be convinced. Keep on pushing :-)

PR #154, opened 2017-02-13
Log creaper commands in JSON format in debug

### LINKED ISSUE #151: Log details of creaper commands

mchoma:
Resolves #151

It producelog output as:
INFO: Executing operation /core-service=management/security-realm=creaperSecRealm:add
JSON format:
{
"operation" : "add",
"address" : [
{
"core-service" : "management"
},
{
"security-realm" : "creaperSecRealm"
}
]
}

### DIFF
--- a/core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineManagementClientImpl.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineManagementClientImpl.java
@@ -1,5 +1,11 @@
 package org.wildfly.extras.creaper.core.online;
 
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
 import org.jboss.as.cli.CliInitializationException;
 import org.jboss.as.cli.CommandContext;
 import org.jboss.as.cli.CommandFormatException;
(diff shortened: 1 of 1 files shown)


mchoma:
following https://github.com/wildfly-extras/creaper/pull/153

Ladicek:
You could have just adjusted #153, but whatever works for you :-)

Ladicek:
Now looking at the 1st comment -- is the JSON format really multiline?

mchoma:
Not sure, if I understand you correctly. When toJSONString(compact=false) is called, then yes operation is printed "nicely" on multiple lines.

Ladicek:
Ah, I definitely don't want that on DEBUG level. That would be too much of a change. And if I remember correctly, you wanted that JSON for pasting into a `curl` invocation, so you should rather want a compact output, no?

mchoma:
Ok. I will add then own TRACE level log entry for JSON format. I prefer compact=false, as it is more human-readable and curl is OK with it.

Ladicek:
Ah, OK, thanks.

Ladicek:
Oh crap, I shouldn't have merged just yet :-( What's the difference between

```java
log.tracef("JSON format:\n%s", operation.toJSONString(false));
```

and

```java
log.tracef("JSON format:%n%s", operation.getOperation().toJSONString(false));
```

?

Specifically, I'm thinking about `\n` vs. `%n`.

Which one is correct?

Ladicek:
@mchoma bump

mchoma:
%n is correct, because it is platform independent.

Should I send new PR with only that change?

rhatlapa:
Yes please.

Ladicek:
Absolutely, thanks.

PR #156, opened 2017-03-06
Add option for ISPN cache removal commands to add the allow-resource-…

jmartisk:
…service-restart header

https://issues.jboss.org/browse/JBEAP-8924 was rejected, as expected :cry: :jack_o_lantern:

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
@@ -1,10 +1,17 @@
 package org.wildfly.extras.creaper.commands.infinispan.cache;
 
+import org.wildfly.extras.creaper.core.ServerVersion;
 import org.wildfly.extras.creaper.core.online.OnlineCommand;
 import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
 import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Headers;
 import org.wildfly.extras.creaper.core.online.operations.Operations;
 
+/**
+ * Removes an Infinispan cache.
+ * Since WildFly 11 / EAP 7.1, this operation requires a reload, and this command
+ * automatically appends the allow-resource-service-restart header.
+ */
 public final class RemoveCache implements OnlineCommand {
     private final Address address;
 
@@ -27,6 +34,8 @@ public final class RemoveCache implements OnlineCommand {
     @Override
     public void apply(OnlineCommandContext ctx) throws Exception {
         Operations ops = new Operations(ctx.client);
+        if (ctx.version.greaterThanOrEqualTo(ServerVersion.VERSION_5_0_0))
+            ops = ops.headers(Headers.allowResourceServiceRestart());
         ops.remove(address);
     }
 }


commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
```diff
@@ -22,11 +28,14 @@ public RemoveCache(String cacheContainer, CacheType cacheType, String cacheName)
...
                 .and("cache-container", cacheContainer)
                 .and(cacheType.getType(), cacheName);
+        this.allowResourceServiceRestart = allowResourceServiceRestart;
     }
 
     @Override
     public void apply(OnlineCommandContext ctx) throws Exception {
         Operations ops = new Operations(ctx.client);
+        if(allowResourceServiceRestart)
+            ops = ops.headers(Values.of("allow-resource-service-restart", "true"));
```
Ladicek: See the `Headers` class.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
```diff
@@ -4,11 +4,17 @@
...
 
 public final class RemoveCache implements OnlineCommand {
     private final Address address;
+    private final boolean allowResourceServiceRestart;
 
     public RemoveCache(String cacheContainer, CacheType cacheType, String cacheName) {
+        this(cacheContainer, cacheType, cacheName, false);
+    }
+
+    public RemoveCache(String cacheContainer, CacheType cacheType, String cacheName, boolean allowResourceServiceRestart) {
```
Ladicek: We don't have a builder for this class yet, right? Bollean arguments are always a good sign that a builder would be a good idea.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
jmartisk: Cool, didn't notice. Will use that.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
jmartisk: Yeah but in this case it sounds a bit of an overkill to me, there are 3 required arguments and one optional, builders are suited for classes with a prevalence of optional arguments :) also for keeping backward compatibility we would need to keep the constructor as well, where it gets even weirder. But if you insist, I can do that...

Ladicek: Yea I see where you're coming from, but the builder can have a constructor with the same 3 arguments and just 1 optional method, so not that big of an overkill IMHO.

Ladicek: Or possibly this could be automatic, if we detect the server isn't able to remove a cache without service restart (version check).

I know this goes against "restarts should be explicit", but it's an option worth considering. WDYT?

jmartisk: Version check sounds good, it will be a bit more difficult to remove a cache without automatically calling service restart, but who does that anyway?

Ladicek: Just please add a comment to the class javadoc that with EAP 7.1+, this command enables the service restart header. That would be good enough from my POV. @rhatlapa, WDYT?

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/infinispan/cache/AddDistributedCacheOnlineTest.java
```diff
@@ -17,6 +17,7 @@
 
 import java.io.IOException;
 import java.util.UUID;
+import java.util.concurrent.TimeoutException;
```
Ladicek: Oh and I wonder, how come this is now necessary? The `apply` method signature didn't change, right?

jmartisk: Right, that's a leftover from my feeble previous attempt where I called `reload`, I'll get rid of that.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
rhatlapa: Shouldn't we try to have the javadoc upstream related => rather pointing out WildFly releases? Otherwise I am OK with both approaches (Builder one and constructor one with one optional argument.
Regarding having it based on Version seems probably better as user usually just wants to remove the cache and have it working on all EAP versions and not need to modify command attributes based on EAP version it is used with.

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/infinispan/cache/AddDistributedCacheOnlineTest.java
Ladicek: Cool, that sounds like this will be a tiny PR in the end, I love those :-)

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
Ladicek: Yea I agree, I always try to use upstream names and versions, unless there's something specific to EAP. So this is only WildFly 11+, I think? Which should be `ServerVersion.VERSION_5_0_0`.

commands/src/main/java/org/wildfly/extras/creaper/commands/infinispan/cache/RemoveCache.java
```diff
@@ -1,10 +1,17 @@
...
+import org.wildfly.extras.creaper.core.ServerVersion;
 import org.wildfly.extras.creaper.core.online.OnlineCommand;
 import org.wildfly.extras.creaper.core.online.OnlineCommandContext;
 import org.wildfly.extras.creaper.core.online.operations.Address;
+import org.wildfly.extras.creaper.core.online.operations.Headers;
 import org.wildfly.extras.creaper.core.online.operations.Operations;
 
+/**
+ * Removes an Infinispan cache.
+ * Since WildFly 11 / EAP 7, this operation requires a reload, and this command
```
Ladicek: EAP 7.1, no?

jmartisk: Whoops! Fixed.

Ladicek:
Thanks!

PR #159, opened 2017-04-20
WildFly 11 support

Ladicek:


testsuite/standalone/src/test/java/org/wildfly/extras/creaper/core/online/HttpsOnlineManagementClientTest.java
```diff
@@ -38,6 +40,9 @@ public static void removeUser() throws CommandFailedException {
     @Before
     @Override
     public void connect() throws IOException {
+        // WildFly 11.0.0.Alpha1 has issues
```
mchoma: what kind of issues is that?

Ladicek: I wish I knew.

Ladicek: I mean:
1. WildFly 11 doesn't know the `http-upgrade-enabled` attribute in XML.
2. These tests fail with `Could not connect to
https-remoting://localhost:9993`, and I have no idea why.

Issue #160, opened 2017-05-24
Neither ReloadToSnapshot nor SnapshotBackup#restore work in domain mode


zebrik:
This is caused by `server-config` parameter being passed to the `reload` command instead of either `domain-config` or `host-config`.

Ladicek:
This is actually very simple: I just don't know enough about domain. If you look at `ReloadToOriginal`, I made a bunch of comments to explain this -- I wasn't careful enough to do the same in `ReloadToSnapshot` / `SnapshotBackup`. And I think there was someone who worked with me on this who also didn't know much about domain :-)

I agree this is important, and I need to ask you for help. How do you think this should work?

zebrik:
The scope of this issue was originally to fix at least snapshots of domain.xml which is IMHO the main use case. To fix just this would be probably easy and I could create PR for this if you want.
The scope of #161 was to consider if we want to extend the `SnapshotBackup` functionality to enable backuping/restoring of host.xml as well. This is probably not so common requirement and would require more work.
Are there any domain tests in Creaper?

Ladicek:
OK, I see. So fixing the backup/restore of domain.xml config would be just a simple change to `ReloadFromSnapshot`, right?

And the ability to backup/restore host.xml config would require more work both in `SnapshotBackup` and `ReloadToSnapshot`, right?

If so, I'm going to reopen #161 and if you provide a PR for this, that would be awesome!

zebrik:
Exactly. I can provide PR for #160 but without test since there are no module for domain tests so far - am I right?

Ladicek:
I always wanted to have a domain testsuite, but never had the courage to actually implement it. So yes, you're right, there's no domain testsuite. YET! :-)

zebrik:
https://github.com/wildfly-extras/creaper/pull/162

Ladicek:
Fixed in #162.

PR #162, opened 2017-05-24
#160: fixing reload to a doman.xml snapshot

zebrik:
See https://github.com/wildfly-extras/creaper/issues/160

### DIFF
--- a/core/src/main/java/org/wildfly/extras/creaper/core/online/operations/admin/ReloadToSnapshot.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/online/operations/admin/ReloadToSnapshot.java
@@ -34,19 +34,19 @@ public final class ReloadToSnapshot {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/wildfly/extras/creaper/core/online/operations/admin/ReloadToSnapshot.java
```diff
@@ -70,4 +70,18 @@ public ModelNodeResult perform(Operations ops, Address address) throws IOExcepti
             return ops.invoke(Constants.RELOAD, address, Values.of(Constants.SERVER_CONFIG, snapshot));
         }
     }
+
+    private static final class ReloadToDomainSnapshotRestartOperation implements RestartOperation {
+        private static final String DOMAIN_CONFIG = "domain-config";
```
Ladicek: Move to the shared `Constants` class.

zebrik:
I have just amended and force pushed requested change.

PR #164, opened 2017-08-03
add Address.profile

simkam:


core/src/main/java/org/wildfly/extras/creaper/core/online/operations/Address.java
```diff
@@ -41,6 +41,10 @@ public static Address extension(String extension) {
         return Address.of(Constants.EXTENSION, extension);
     }
 
+    public static Address profile(String profile) {
+        return Address.of(Constants.PROFILE, profile);
+    }
+
```
Ladicek: Care to update the class' javadoc? :-) I know it's missing other methods too, so I was lazy during previous reviews, but this time, I actually remembered about this :-)

simkam:
added missing javadoc

Ladicek:
Thanks!

PR #165, opened 2018-03-15
Added wildfly12 profile

olukas:
Added wildfly12 profile and activate it by default for tests.

Implementation of `executeAsync` in `HttpModelControllerClient` was added due to changes in WildFly Core in https://github.com/wildfly/wildfly-core/commit/4713ebe9a1305be21f72754aa7634e971156a360#diff-bb461a3456601632260a50ddc3cc462cR628.

### DIFF
--- a/core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
@@ -28,14 +28,22 @@ import org.jboss.as.controller.client.OperationMessageHandler;
 import org.jboss.as.controller.client.OperationResponse;
 import org.jboss.dmr.ModelNode;
 import org.jboss.threads.AsyncFuture;
+import org.jboss.threads.AsyncFutureTask;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 /**
  * Controller provides execution of {@link ModelNode} or {@link Operation} over HTTP.
- * Asynchronous execution is not provided. Controller does not inform about any progress, therefore, there is no point
- * in using {@link OperationMessageHandler}.
+ * Controller does not inform about any progress, therefore, there is no point in using {@link OperationMessageHandler}.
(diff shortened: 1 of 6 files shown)


olukas:
Failed test cases (`JournalStoreVsJdbcStoreOnlineTest` and `ChangeJdbcAttributesOnlineTest`) in Travis CI are not caused by this PR. I reported and issue https://github.com/wildfly-extras/creaper/issues/166 for those test cases.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -124,28 +134,28 @@ public OperationResponse executeOperation(Operation operation, OperationMessageH
         return OperationResponse.Factory.createSimple(execute(operation));
     }
 
-    /**
-     * <b>Not supported!</b>
-     */
     @Override
     public AsyncFuture<ModelNode> executeAsync(ModelNode modelNode, OperationMessageHandler handler) {
-        throw new UnsupportedOperationException("Asynchronous execution is not supported by " + getClass().getName());
+        ExecutorService executor = Executors.newSingleThreadExecutor();
```
Ladicek: When is the executor shut down? I guess the `...AsyncFutureTask` doesn't shutdown the executor after it's completed, so you'll keep leaking threads.

Given that this class has a `close` method, maybe it would be better to have one `Executors.newCachedThreadPool()` executor for the entire instance of this class, which you could shutdown in the `close` method?

olukas: Thanks for advice. Calling `shutdown()` was missing. Use only one `ExecutorService` for entire instance is better solution. I fixed it in suggested way.

core/src/main/java/org/wildfly/extras/creaper/core/online/HttpModelControllerClient.java
```diff
@@ -124,33 +135,43 @@ public OperationResponse executeOperation(Operation operation, OperationMessageH
...
 
     @Override
     public void close() throws IOException {
         client.close();
+        if (executorService != null) {
+            executorService.shutdown();
+        }
+    }
+
+    private ExecutorService obtainExecutorService() {
```
Ladicek: I don't think the executor service needs to be created lazily; if the client is created, it's extremely likely to be used :-) But I'll leave it to you -- this is good too.

olukas: I created it lazily because older version than WildFly 12 still uses `execute` method instead of `executeAsync`. I believe it is better to created ExecutorService only on demand.

Ladicek: Ha, very much true! Makes sense.

Issue #166, opened 2018-03-15
Tests ChangeJdbcAttributesOnlineTest and JournalStoreVsJdbcStoreOnlineTest fail on wildfly11 and higher


olukas:
Tests ChangeJdbcAttributesOnlineTest and JournalStoreVsJdbcStoreOnlineTest fail on wildfly11 and higher due to "WFLYCTL0105: use-journal-store is invalid in combination with use-jdbc-store". This restriction should be reflected in given commands.

olukas:
Also `ChangeBasicAttributesOnlineTest` rarely fails with the same exception.

simkam:
test updated in #169

simkam:
reopening, unfortunately #169 doesn't fix this

PR #169, opened 2018-03-27
JournalStoreVsJdbcStoreOnlineTest fix

ChesterSVK:
Fix for issue: https://github.com/wildfly-extras/creaper/issues/166 

Use-journal-store and use-jdbc-store attributes must be set exclusively and server must be restarted for avoiding this conflict.


olukas:
+1 for check in builder whether only one of journal store and JDBC store is configured. However I think that disabling JDBC when enabling journal and vice versa can cause some tricky issues for users - e.g. in case when they accidentally try to configure both store - in that case I believe that just thrown Exception is better than use last configured value for store.

@ChesterSVK As I said in https://github.com/wildfly-extras/creaper/issues/166 - `ChangeJdbcAttributesOnlineTest` and `ChangeBasicAttributesOnlineTest` intermittently fail with the same failure-description. Could you please fix also those test cases as part of this PR?

Ladicek:
> However I think that disabling JDBC when enabling journal and vice versa can cause some tricky issues for users

Yea, should probably just document that WildFly changed behavior in this.

olukas:
@ChesterSVK Hi, is there any progress with fixing `ChangeJdbcAttributesOnlineTest` and `ChangeBasicAttributesOnlineTest` please?

rsvoboda:
@msve please take a look

@Ladicek can you grant rights for creaper repo to @msve ?
He is "component owner" now on eap qe side.

PR #170, opened 2018-04-04
Use explicit location in tests which use AddCredentialStore

olukas:
Since WFLY 13 location for AddCredentialStore is required when JCEKS type is used. In previous version WildFly is able to use some magic and generate location with name derived from resource name. However for Creaper command test purposes it is safer to use some named location.

### DIFF
--- a/testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/elytron/credentialstore/AddCredentialStoreOnlineTest.java
+++ b/testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/elytron/credentialstore/AddCredentialStoreOnlineTest.java
@@ -8,12 +8,17 @@ import java.io.IOException;
 import org.jboss.arquillian.junit.Arquillian;
 import org.junit.After;
 import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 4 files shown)


testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/elytron/credentialstore/AddCredentialStoreOnlineTest.java
```diff
@@ -24,13 +29,26 @@
     private static final String TEST_CREDENTIAL_STORE_NAME2 = "CreaperTestCredentialStore2";
     private static final Address TEST_CREDENTIAL_STORE_ADDRESS2 = SUBSYSTEM_ADDRESS
             .and("credential-store", TEST_CREDENTIAL_STORE_NAME2);
+    private static final String PATH_ADDRESS_NAME = "tmp.path";
+    private static final Address TMP_PATH_ADDRESS = Address.of("path", PATH_ADDRESS_NAME);
 
     private static final String TEST_PASSWORD = "somePassword";
 
+    private static String tmpFilePath;
```
mchoma: why is this necessary? and TemporaryFolder tmp is not enough?

mchoma: Is this just shortcut for tmp.getRoot().getAbsolutePath() ?

olukas: yes, I want to avoid multiple `tmp.getRoot().getAbsolutePath()`

Issue #172, opened 2018-04-25
Creaper doesn't compile with Java 9


rsvoboda:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.15:checkstyle (check-style) on project creaper-parent: Execution check-style of goal org.apache.maven.plugins:maven-checkstyle-plugin:2.15:checkstyle failed: Plugin org.apache.maven.plugins:maven-checkstyle-plugin:2.15 or one of its dependencies could not be resolved: Could not find artifact com.sun:tools:jar:1.6.0 at specified path /Library/Java/JavaVirtualMachines/jdk-9.0.4.jdk/Contents/Home/../lib/tools.jar -> [Help 1]

Tools jar is not available in Java 9+

rsvoboda:
Something to consider is dependency on https://github.com/jboss/jboss-parent-pom to solve plugin dependencies and Java 9+ related issues

jstourac:
no idea, main purpose of that was to get rid of JDK6 and JDK7 to ease process of a new version release :) I may check once I have a bit of spare time.

simkam:
Creaper now compiles with JDK11, CI (GH Actions) uses JDK 8 and JDK 11 now.

Issue #157, opened 2017-04-12
investigate moving to Java 8


Ladicek:
A lot of Creaper dependencies can't be upgraded because Creaper is still on Java 6. Also Creaper depends on Guava, but it only uses a couple of utils that are provided by Java 8 natively. Also Creaper can't use lambdas and default methods... etc. etc.

This would of course be 2.0 thing.

A separate possibility would be dropping support for EAP 6 and WildFly 8/9, but I'm not proposing that. EAP officially supports Java 8 since 6.4, so that would be fine; question is if moving Creaper to Java 8 wouldn't break some EAP QE processes. Need to figure out.

simkam:
https://github.com/wildfly-extras/creaper/pull/193 moves build to JDK8

simkam:
done in #193 (there are still some pending tasks like documentation updates)

PR #175, opened 2018-10-25
move build to JDK 1.8, support JDK11

simkam:
I think it's time to move to 2.0. This is mainly to start discussion. 

touches #172 and #157 

It fixes only build (mvn clean install -DskipTests), there are no changes to test suite for now. 

I tried to make version compatible with 1.6+, but due to changes mainly to checkstyle and findbugs it is very hard.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -486,7 +486,7 @@
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-javadoc-plugin</artifactId>
                         <configuration>
-                            <additionalparam>-Xdoclint:none</additionalparam>
+                            <doclint>none</doclint>
                         </configuration>
                     </plugin>
                 </plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,6 @@
         <version.org.apache.maven.plugins.maven-surefire-plugin>2.22.1</version.org.apache.maven.plugins.maven-surefire-plugin>
         <version.org.bouncycastle>1.60</version.org.bouncycastle>
         <version.org.codehaus.groovy.groovy-everything>2.5.3</version.org.codehaus.groovy.groovy-everything>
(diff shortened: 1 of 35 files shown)


pom.xml
```diff
@@ -453,33 +437,26 @@
...
+                <artifactId>spotbugs-maven-plugin</artifactId>
             </plugin>
         </plugins>
     </build>
 
     <profiles>
-        <!-- support Java 8: use newer FindBugs plugin, switch off Javadoc's doclint -->
+        <!-- switch off Javadoc's doclint -->
         <profile>
             <id>jdk8</id>
```
Ladicek: I'd get rid of this profile and fold it into the main config.

pom.xml
```diff
@@ -403,30 +408,9 @@
                     </configuration>
                 </plugin>
                 <plugin>
-                    <groupId>org.codehaus.mojo</groupId>
-                    <artifactId>codenarc-maven-plugin</artifactId>
```
Ladicek: What's wrong with CodeNarc? It doesn't work with 8?

pom.xml
simkam: good point, it's leftover from times I thought that it would work with older JDKs. I'll move it to main config.

pom.xml
simkam: it doesn't work with JDK 11, I'll keep it with JDK8.

it uses by default codenarc 0.22 and groovy-all 1.7.something.
Groovy 1.7.x isn't compatible with JDK11.
Latest versions of codenarc is 1.2.1 and of groovy-all is 2.5.3.
Both is configurable with codeNarcVersion and groovyVersion, but groovy-all download fails because it's pom type in this version.

codenarc-maven-plugin development doesn't seem to active since 2015

rsvoboda:
@msve please take a look

@Ladicek can you grant rights for creaper repo to @msve ?
He is "component owner" now on eap qe side.

simkam:
I'm closing this for now. It needs updates to current versions of components and retesting with latest JDKs.

PR #185, opened 2020-10-06
Add Wildfly 15 - 20 versions.

jstourac:
FYI @simkam

### DIFF
--- a/core/src/main/java/org/wildfly/extras/creaper/core/ServerVersion.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/ServerVersion.java
@@ -76,6 +76,14 @@ public final class ServerVersion {
     public static final ServerVersion VERSION_7_0_0 = new ServerVersion(7, 0, 0);
     /** WF 14.0.x.Final */
     public static final ServerVersion VERSION_8_0_0 = new ServerVersion(8, 0, 0);
+    /** WF 15.0.x.Final */
+    public static final ServerVersion VERSION_9_0_0 = new ServerVersion(9, 0, 0);
+    /** WF 16.0.x.Final, 17.0.x.Final and 18.0.x.Final */
+    public static final ServerVersion VERSION_10_0_0 = new ServerVersion(10, 0, 0);
+    /** WF 19.0.x.Final */
+    public static final ServerVersion VERSION_12_0_0 = new ServerVersion(12, 0, 0);
+    /** WF 20.0.x.Final */
+    public static final ServerVersion VERSION_13_0_0 = new ServerVersion(13, 0, 0);
 
     private static final ServerVersion[] KNOWN_VERSIONS = {
             VERSION_0_0_0,
@@ -99,6 +107,10 @@ public final class ServerVersion {
             VERSION_6_0_0,
             VERSION_7_0_0,
             VERSION_8_0_0,
+            VERSION_9_0_0,
+            VERSION_10_0_0,
+            VERSION_12_0_0,
+            VERSION_13_0_0
     };
 
     /**
--- a/README.md
+++ b/README.md
@@ -267,6 +267,144 @@ Dependencies for `creaper-commands`:
         <version>6.0.2.Final</version>
     </dependency>
 
+#### WildFly 15 (15.0.0) (based on WildFly Core 7)
+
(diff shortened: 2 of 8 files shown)


core/src/main/java/org/wildfly/extras/creaper/core/ServerVersion.java
```diff
@@ -76,6 +76,14 @@
     public static final ServerVersion VERSION_7_0_0 = new ServerVersion(7, 0, 0);
     /** WF 14.0.x.Final */
     public static final ServerVersion VERSION_8_0_0 = new ServerVersion(8, 0, 0);
+    /** WF 15.0.x.Final */
+    public static final ServerVersion VERSION_9_0_0 = new ServerVersion(9, 0, 0);
+    /** WF 16.0.x.Final, 17.0.x.Final and 18.0.x.Final */
+    public static final ServerVersion VERSION_10_0_0 = new ServerVersion(10, 0, 0);
```
jstourac, line 82: @simkam Could you please confirm I didn't make any mistake here? I checked the version number that WildFly 16-18 returns via CLI - `ls` in root dir.

Ladicek: Is `11` really not used anywhere? In some EAP minor version perhaps? (Genuine question, I really don't know.)

jstourac: Well, I didn't find it anywhere:

JBoss EAP `7.2.x` - 8.0.0
JBoss EAP `7.3.x` - 10.0.0
JBoss EAP `7.4.0.CD18` - 10.0.0
JBoss EAP `7.4.0.CD19` - 12.0.0
JBoss EAP `7.4.0.CD20` - 13.0.0

Also, changing major management version in minor server update would not be possible/allowed anyway, right?

What I find confusing is that now three WildFly versions cannot be distinguished by management version even though, there seems to be at least one change in management between WildFly 17 and WildFly 18 since new attribute `/subsystem=elytron/syslog-audit-log[reconnect-attempts]` was added in the latter version. I ran into this change by accident so maybe there are even more differences. Looks like there was an issue in version number propagation...?

Ladicek: > Also, changing major management version in minor server update would not be possible/allowed anyway, right?

That's right, but something like that (management version change in EAP micro) already happened once :-)

> What I find confusing is that now three WildFly versions cannot be distinguished by management version

Yea agree. Too bad :-( And probably a bug. Better to raise with WildFly devs, I guess.

jstourac: Well, not sure whether it makes sense to raise a bug since latest WildFly final release available is 20 and beta of WildFly 21 is also available already.

I guess there is not anything we can do about this anymore anymore, or do I miss something?

Ladicek: They might at least be able to provide some insight. Whether this was intentional, or actual bug. No, I wouldn't expect it to be fixed retroactively :-)

simkam: I checked git history of multiple branches in wildfly-core and wildfly-core-eap and version 11 really doesn't seem to be used anywhere. I'm not going to dig into it. But feel free to do so @jstourac.

jstourac: I probably won't. Just for possible future reference, there are following two links in regards of upgrade from `10.0.0` to `12.0.0`:

* https://issues.redhat.com/browse/WFCORE-4822
* https://github.com/wildfly/wildfly-core/commit/d9c0a6b0682af1fa0d460046bf23f7c594a4572b#diff-7365e20680bb2d70249250345e3953c6

simkam: I asked on chat

> I think for that core release no modifications were made at all but the management version was assumed to be closely aligned with the .Final version for things like host excludes so it was left unused.

simkam (approved):

PR #186, opened 2020-10-07
Add support for `cipher-suite-names` of `*-ssl-context` element.

jstourac:
This cannot be merged until this MR https://github.com/wildfly-extras/creaper/pull/185 is merged first.

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/elytron/CreateServerSSLContext.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/elytron/CreateServerSSLContext.java
@@ -32,6 +32,7 @@ public final class CreateServerSSLContext implements OnlineCommand {
     private final String name;
     protected final List<String> protocols;
     private final String cipherSuiteFilter;
+    private final String cipherSuiteNames;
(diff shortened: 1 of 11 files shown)


jstourac:
@simkam rebased

commands/src/main/java/org/wildfly/extras/creaper/commands/elytron/CreateServerSSLContext.java
```diff
@@ -62,6 +63,10 @@
     // Multiple usage
     private final String algorithm;         // keystore manager, truststore manager
 
+    // Default set of cipher suites for TLSv1.3 to be set in 'cipher-suite-names' attribute.
+    public static final String CIPHER_SUITE_NAMES_VALUE =
```
simkam, line 67: I wonder if we could find some better name for this constant, for example `TLS13_CIPHER_SUITE_NAMES` ?

jstourac: Okay, makes sense, I've just updated. Thanks for this! :)

simkam (approved):

PR #193, opened 2021-10-02
Move to JDK8 as a minimum required version

jstourac:
Let's ditch the JDK6 and the JDK7 support and move to the JDK8 so we can possibly update some plugins and dependencies, move our code to a more modern one but mostly - to ease the possibility of performing a new release of Creaper :)

Main point of this MR is not support of any JDK11+ now.

Partly solves #157 - following things will be left to be done in that issue:

* get rid of the Guava dependency
* consider the EAP6 and the WildFly 8/9 support removals

Slightly inspired by #175.


mchoma:
Thanks for that. Overall looks good/safe to me. I am not following this task in deep. Just regarding checklist from description:

- get rid of the Guava dependency
  - I see it still there
- consider the EAP6 and the WildFly 8/9 support removals
  - I see it was not. However we can in version 2.0 if this helps. There is very small chance we will need creaper update for EAP6. If it happens we can release from branch derived from `v1.6.2` tag.

simkam (approved):

simkam (approved):

PR #198, opened 2021-10-12
Removal of AS7 and WildFly 8/9 support

jstourac:
There may be some topic for discussion - default protocol for remoting + default port number.

I kept the default to AS7, which is `remoting` and `9999` but in WildFly the default is `http-remoting` and `9990`. Reason to keep that is that I was too lazy to change this. But if you feel that we should change this behavior (including the `creaper.wildfly` property), I shall update this MR appropriately.


simkam (changes requested):
can you add these changes to changelog?

jstourac:
@simkam thank you, updated

simkam (approved):

PR #208, opened 2022-06-16
add WildFly 23

simkam:

depends on #206 and #207


testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/elytron/credentialstore/AddCredentialStoreOnlineTest.java
```diff
@@ -48,8 +48,8 @@ public static void createTmpFilePath() throws IOException {
     public void cleanup() throws Exception {
         ops.removeIfExists(TEST_CREDENTIAL_STORE_ADDRESS2);
         ops.removeIfExists(TEST_CREDENTIAL_STORE_ADDRESS);
-        ops.removeIfExists(TMP_PATH_ADDRESS);
         administration.reloadIfRequired();
+        ops.removeIfExists(TMP_PATH_ADDRESS);
```
jstourac, line 52: I'm thinking that we should have another `administration.reloadIfRequired()` call here again to finish the cleanup task?

jstourac: well, technically we should but since the deleting operation over the `/path` shouldn't require that, we can probably leave it as is now...

jstourac (approved):

PR #209, opened 2022-06-16
add WildFly 24

simkam:

depends on #206, #207 and #208


jstourac:
I've created #213 to tracking that we want to re-enable bits that are commented out now.

jstourac (approved):

PR #210, opened 2022-06-16
add WildFly 25

simkam:

depends on #206, #207, #208 and #209

### DIFF
--- a/commands/src/main/java/org/wildfly/extras/creaper/commands/security/AddAuthorizationModule.java
+++ b/commands/src/main/java/org/wildfly/extras/creaper/commands/security/AddAuthorizationModule.java
@@ -7,6 +7,7 @@ import java.util.concurrent.TimeoutException;
 import org.wildfly.extras.creaper.commands.foundation.offline.xml.GroovyXmlTransform;
 import org.wildfly.extras.creaper.commands.foundation.offline.xml.Subtree;
 import org.wildfly.extras.creaper.core.CommandFailedException;
+import org.wildfly.extras.creaper.core.ServerVersion;
 import org.wildfly.extras.creaper.core.offline.OfflineCommand;
 import org.wildfly.extras.creaper.core.offline.OfflineCommandContext;
 import org.wildfly.extras.creaper.core.online.CliException;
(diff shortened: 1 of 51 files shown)


testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/security/AddAuthorizationModuleOnlineTest.java
```diff
@@ -44,6 +47,15 @@ public class AddAuthorizationModuleOnlineTest {
...
             .and("policy-module", TEST_AUTHORIZATION_MODULE_NAME_2);
 
+    @BeforeClass
+    public static void checkServerVersionIsSupported() throws Exception {
+        // check version is supported
+        ServerVersion serverVersion
+                = ManagementClient.online(OnlineOptions.standalone().localDefault().build()).version();
+        Assume.assumeFalse("Legacy security was removed in WildFly 25.",
+                serverVersion.greaterThanOrEqualTo(ServerVersion.VERSION_18_0_0));
+    }
```
jstourac, line 57: Just a proposal - wouldn't be better to create some common function for this and use it in all relevant places instead of having the very same code on multiple places? :angel:

commands/src/main/java/org/wildfly/extras/creaper/commands/security/AddAuthorizationModule.java
```diff
@@ -43,6 +44,10 @@ private AddAuthorizationModule(Builder builder) {
     @Override
     public void apply(OnlineCommandContext ctx) throws CliException, CommandFailedException, IOException,
             TimeoutException, InterruptedException {
+        if (ctx.version.greaterThanOrEqualTo(ServerVersion.VERSION_18_0_0)) {
+            throw new AssertionError("Legacy security was removed in WildFly 25.");
+        }
```
jstourac, line 49: Again, having some common function would be probably nicer. But I don't insist.

jstourac:
Just two comments but LGTM otherwise :)

testsuite/standalone/src/test/java/org/wildfly/extras/creaper/commands/security/AddAuthorizationModuleOnlineTest.java
simkam: I was thinking about it but in the end, I followed the same convention as it is in elytron commands, there is a similar check in every elytron command.

jstourac (approved):

PR #216, opened 2023-01-03
don't use deprecated method to create ModelControllerClient

simkam:
- Already deprecated in WildFly Core 2.0.10.Final / WildFly 10.
- Removed in WildFly Core 20.0.0.Beta3 / to be in WildFly 28.

Also removed some legacy code for AS7 support, we dropped support for AS7 in Creaper 2.0.0-Alpha.1.

I'm running tests with WildFly 10 locally, I'll comment with results. GH Actions will test WildFly 24 (we haven't proper support for > 24 yet).


simkam:
tests passed locally with WildFly 10

jstourac (approved):

Issue #199, opened 2021-10-13
`creaper.wildfly` property + `remoting` protocol and `9999` port defaults...


jstourac:
For AS7, there was a default protocol used e.g. for CLI connection named `remoting` with `9999` port number as default. Since WildFly 9, this has changed to `http-remoting` or `https-remoting` and default port number `9990` or `9993` respectfully. For the compatibility reason the original default could be easily changed by providing `creaper.wildfly` property.

We shall probably remove that property completely since we support only WildFly 10+ now and use `http-remoting` with `9990` port as default.

JFTR https://github.com/wildfly-extras/creaper/pull/198

PR #219, opened 2023-01-09
[#199] default to WildFly connection configuration

simkam:
fixes https://github.com/wildfly-extras/creaper/issues/199

- switch default protocol to `http-remoting`
- switch default port to 9990
- drop `creaper.wildfly` property
- drop `remoting` protocol support

### DIFF
--- a/core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
+++ b/core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
@@ -23,7 +23,6 @@ import org.wildfly.extras.creaper.core.ManagementClient;
  * quite a number of public types.
  */
 public final class OnlineOptions {
-    private static final String CREAPER_WILDFLY = "creaper.wildfly";
 
     public final boolean isStandalone;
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/wildfly/extras/creaper/core/online/OnlineOptions.java
```diff
@@ -196,10 +191,8 @@ private ConnectionOnlineOptions(Data data) {
 
         /**
          * <p>Connect to {@code localhost} and use the default management port of the application server.
-         * This is {@code 9999} by default (JBoss AS 7), and if the system property {@code creaper.wildfly} is defined,
-         * it changes to 9990 or (if {@link OptionalOnlineOptions#ssl(SslOptions) ssl} is used) to 9993 (WildFly).
-         * This makes it easy to run the same code against both AS7 and WildFly only by defining a single system
-         * property.</p>
+         * This is {@code 9990} by default, but if {@link OptionalOnlineOptions#ssl(SslOptions) ssl} it is 9993.
```
jstourac, line 194:
```suggestion
         * This is {@code 9990} by default, but if {@link OptionalOnlineOptions#ssl(SslOptions) ssl} is not null, then it is 9993.
```

testsuite/pom.xml
```diff
@@ -1125,7 +1123,7 @@
                 <arquillianContainer.version>${wildfly25.arquillianContainer.version}</arquillianContainer.version>
 
                 <tests.ignoredCategory>${ignoredCategory.wildfly11}</tests.ignoredCategory>
-                <tests.specialJvmArgs>${specialJvmArgs.wildfly}</tests.specialJvmArgs>
+                <tests.specialJvmArgs></tests.specialJvmArgs>
```
jstourac, line 1126: well, we could probably delete these all `specialJvmArgs` properties but whatever...

jstourac:
LGTM, thank you - just one actual small concern regarding one comment line.

testsuite/pom.xml
simkam: let's keep it for now, if we ever need it. It might be useful for JDK17 for example.

PR #235, opened 2024-09-02
Update AddDataSource offline command based on WFLY-18324

marekkopecky:
[WFLY-18324](https://issues.redhat.com/browse/WFLY-18324) introduce new user-name and password attributes and drop user-name and password elements in configuration.

security-domain element in security element hasn't been changed, see latest xsd [here](https://github.com/wildfly/wildfly/blob/main/connector/src/main/resources/schema/wildfly-datasources_7_2.xsd#L795).

This PR changes AddDataSource creaper command based on the latest changes.


marekkopecky:
@simkam Can you please check this PR?

simkam:
We need to use the correct command depending on the server version
```
if (serverVersion.lessThan(ServerVersion.VERSION_X_X_X)) {
  use elements
} else {
  use attributes
}
```
The problem is that our last version in creaper is WildfFly 28. There is WIP https://github.com/wildfly-extras/creaper/pull/232, but we need at least also 30 because [WFLY-18324](https://issues.redhat.com/browse/WFLY-18324) was fixed there.

simkam:
@marekkopecky we have added WF30 support, can you please rebase this and include condition mentioned above, WF30 management version is `VERSION_23_0_0`.

marekkopecky:
@simkam  Thank you for your feedback! Do you know how creaper parse server version? I'm using WF from master branch (34.0.0.Beta1-SNAPSHOT which is using wf-core 26.0.0.Beta2), but creaper reports server-version 20.0.0 ...

simkam:
Hmm, Creaper relies on the root namespace version in the configuration file and expects it to be the same as the management version. It might be a bug in WildFly; let's take a closer look.

marekkopecky:
Ok, got it, thanks!

management versions:

```
$ ./jboss-cli.sh -c "--command=:read-resource" | grep management.*version
        "management-major-version" => 27,
        "management-micro-version" => 0,
        "management-minor-version" => 0,
```

namespace version:

 ```
$ cat standalone.xml | grep server.*xmlns
<server xmlns="urn:jboss:domain:20.0">
```

marekkopecky:
@simkam The change from [WFLY-18324](https://issues.redhat.com/browse/WFLY-18324) has been implemented in WF30. WF29 has the same xmlns as wf34: `<server xmlns="urn:jboss:domain:20.0">`. So xmlns of server element can't be used for differentiate different datasource security definition. Is this a bug in WF? I can report product jira for it ...

marekkopecky:
Anyway, I reverted latest commit, because it doesn't seem to be working. Please let me know if you prefer another effort on this PR ...

marekkopecky:
I added a workaround for a detection of datasource subsystem version, as we discussed on chat

simkam (approved):

simkam:
Thank You!

PR #237, opened 2024-09-05
Resolves#236 Update the namespace detection for the root namespace to support stability level qualifiers.

darranl:
Within the regular expression I did first try an extra segment of lowercase alpha characters followed by the colon but that is where I saw we had an overlap with the namespaces of some of our early subsystems so instead specify the known qualifiers.

Although they are not all in use at the moment and which one is present by default will change over time I added tests for all stability levels to check they can all be parsed correctly.

### DIFF
--- a/core/src/test/java/org/wildfly/extras/creaper/core/offline/OfflineServerVersionTest.java
+++ b/core/src/test/java/org/wildfly/extras/creaper/core/offline/OfflineServerVersionTest.java
@@ -68,6 +68,10 @@ public class OfflineServerVersionTest {
     private static final String WFLY27_ROOT = "20.0";
     private static final String WFLY28_ROOT = "21.0";
 
+    private static final String COMMUNITY = "community";
+    private static final String PREVIEW = "preview";
+    private static final String EXPERIMENTAL = "experimental";
+
     @Rule
     public final TemporaryFolder tmp = new TemporaryFolder();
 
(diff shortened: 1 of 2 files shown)


darranl:
FYI this is the WildFly Core PR that exposed this issue:

https://github.com/wildfly/wildfly-core/pull/6152

simkam (approved):

core/src/test/java/org/wildfly/extras/creaper/core/offline/OfflineServerVersionTest.java
```diff
@@ -296,6 +317,23 @@ public void discoverHostXml_wfly28() throws IOException {
         test(ServerVersion.VERSION_21_0_0, HOST_XML, WFLY28_ROOT, EAP7_LOGGING, EAP8_EE);
     }
 
+    @Test
+    public void discoverHostXml_wfly34_community() throws IOException {
+        test(ServerVersion.VERSION_20_0_0, HOST_XML, COMMUNITY, WFLY27_ROOT, EAP7_LOGGING, EAP8_EE);
```
jbliznak, line 322: Just note: the version of root element does not correspond to the test target but we will be updating supported WF versions in next PR so might fix it there.

darranl: +1 there is no guarantee that the schema version will even be updated on a major WildFly version so many WildFly versions will share the same schema version even if they have distinct model versions.

darranl:
@jbliznak Thank you for the suggestions, I committed them to the topic branch.

jbliznak (approved):

simkam:
Thank You!

Issue #244, opened 2024-09-16
Deal with wildfly-core upgrade to jboss-logging 3.6.x


simkam:
After [WFCORE-6883](https://issues.redhat.com/browse/WFCORE-6883), wildfly-core was updated to depend on the new `jboss-logging` version, which comes with some new methods. At the same time, Wildly Integration was updated to use those new methods.

Because of that, we started to see many issues like:
```
java.lang.NoSuchMethodError: 'java.lang.Object org.jboss.logging.Logger.getMessageLogger(java.lang.invoke.MethodHandles$Lookup, java.lang.Class, java.lang.String)' at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:365) at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:273) at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:238) at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:159)
```

or
 
```
[...]

PR #245, opened 2024-09-16
[#244] drop explicit dependency on jboss-logging

simkam:
fixes https://github.com/wildfly-extras/creaper/issues/244

*this is candidate for minor Creaper version bump to 2.1*

It attempts to fix issue due to current dependecy situation:
* creaper has explicit dependency on jboss-logging 3.3.0.Final
* jboss-logging added new methods in 3.6.0
* wildfly-core uses new methods from jboss-logging 3.6.x since 26.0.0.Beta3 which means WildFly 34
* it results in #244

Let's bring in jboss-logging transitivitely from wildfly-core to keep version aligned.

----

Originally jboss-logging dependency was added with this commit message

```
OnlineManagementClient and OfflineManagementClient now log all commands
and operations they are performing. Operations (online only) are logged
on the DEBUG level, commands are logged on INFO. Also, commands are now
advised to provide a toString() method that is used for logging.

The logging framework that is used in Creaper is JBoss Logging, because:

1. It is not actually a logging framework, it's only a facade that delegates
   the logging work to some logging framework that appears to be used.
2. It can detect all commonly used logging frameworks (Logback, Log4j, JUL)
   and one uncommon too (JBoss LogManager).
3. Creaper already depended on it transitively. Avoiding new dependencies
   is good.
4. It has a really small and nice API.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,10 @@ __You have to provide both `*-controller-client` and `*-cli` yourself.__
 If you want to use commands for patching, you also have to provide
 `wildfly-patching`.
 
+You might need to tweak other dependencies to transitively bring
+compatible `jboss-logging` from `wildfly-core` or explicitly provide
+compatible version. WildFly 34 and newer requires at least `3.6.0`.
+
 Other dependencies might be required if you're using "local" CLI operations,
 such as `jms-queue add ...` on WildFly 10 (it was a built-in operation
 in previous WildFly versions, but was moved to a separate module due to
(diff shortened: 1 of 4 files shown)


jamezp (approved):

jamezp (approved):

README.md
```diff
@@ -68,6 +68,10 @@ __You have to provide both `*-controller-client` and `*-cli` yourself.__
 If you want to use commands for patching, you also have to provide
 `wildfly-patching`.
 
+You might need to tweak other dependencies to transitively bring
+compatible `jboss-logging` from `wildfly-core` or explicitly provide
+compatible version. Wildfly 34 and newer requires at least `3.6.0`.
```
jbliznak, line 73:
```suggestion
compatible version. WildFly 34 and newer requires at least `3.6.0`.
```

jbliznak (approved):

